PR #4, opened 2020-07-14
Handle partiality flag in M/ISYM to improve support for handling unmerged data

### LINKED ISSUE #3: DataSet.hkl_to_observed() does not handle partiality flag in M/ISYM
Currently, methods that create or use `M/ISYM` columns do not account for partiality flags. This occurs in `DataSet.hkl_to_observed()`, where the ISYM value is used to map Miller indices, but the `M/ISYM` column is then left intact (though it no longer applies to the mapped indices). Here's a short example using `HEWL_unmerged.mtz` in `tests/data/algorithms`:

```python
mtz = rs.read_mtz("HEWL_unmerged.mtz")
print(mtz["M/ISYM"])
```
outputs:
```
H    K    L  
-22  -9    4      5
-7    18  -9      8
 19  -26   15     7
-15   19   4      3
 15  -2    9     14
                 ..
-3   -17   6     16
-19  -24   1     16
-33   5    6     10
-5   -22   4     16
 29   10   14     1
Name: M/ISYM, Length: 20597, dtype: M/ISYM
```

Instead, this function should map the Miller indices (which it does currently), and extract the partiality flag into a new column. 

Similarly, `DataSet.hkl_to_asu()` should take such a partiality flag into account, if it exists, in order to write a correct `M/ISYM` column. I think it also makes sense to make sure that any IO methods can handle such a column describing partiality method to make sure that one can read/write unmerged reflection data without loss of information.

JBGreisman:
This PR closes #3 . Adds support for use of PARTIAL column to store boolean flag based on M/ISYM column. This flag is set through calls to `DataSet.hkl_to_observed()`, and is then removed by `DataSet.hkl_to_asu()` when the flag is incorporated into setting the M/ISYM column.


JBGreisman:
To Do:
- [ ] Update docstrings to describe cases where unmerged and merged `DataSet` objects have different behavior
- [ ] Add documentation section describing special column names for handling `CENTRIC` and `PARTIAL` flags and their behavior in `DataSet.write_mtz()`

reciprocalspaceship/dataset.py
```diff
@@ -539,16 +548,21 @@ def hkl_to_observed(self, m_isym="M/ISYM", inplace=False):
             dataset = self.copy()
 
         hkls = dataset.get_hkls()
-        isym = (dataset[m_isym] % 256).to_numpy() # Convert M/ISYM to ISYM
-        observed_hkls = hkl_to_observed(hkls, isym, dataset.spacegroup)
 
-        # Update HKLs
+        # GH#3: Separate combined M/ISYM into M and ISYM
+        isym = (dataset[m_isym] % 256).to_numpy()
```
kmdalton, line 553: can we change the default value of `m_isym` to `None`, and check the `self.dtypes` for any `rs.dtypes.m_isym`? we will need to raise some kind of error in the case that multiple columns have `rs.dtypes.m_isym`.

reciprocalspaceship/dataset.py
```diff
@@ -539,16 +548,21 @@ def hkl_to_observed(self, m_isym="M/ISYM", inplace=False):
...
+        
+        # Update HKLs        
+        observed_hkls = hkl_to_observed(hkls, isym, dataset.spacegroup)
         index_keys = dataset.index.names
         dataset.reset_index(inplace=True)
         dataset[["H", "K", "L"]] = observed_hkls
         dataset[["H", "K", "L"]] = dataset[["H", "K", "L"]].astype("HKL")
         dataset.set_index(index_keys, inplace=True)
-
+        dataset.sort_index(inplace=True)
```
kmdalton, line 564: Is there a reason we need to sort the indices here? Preserving the order leaves the option open for copying over columns from a version of the `Dataset` that has a numerical index (I think). This may not be a very important use case, but I wanted to make sure this won't break anything.

JBGreisman: I agree this sort is unnecessary -- I'll remove it

reciprocalspaceship/dataset.py
JBGreisman: I agree. This will be handled with an analogous method to `DataSet.get_phase_keys()`.

JBGreisman:
In addition to improving handling of M/ISYM column names, the above commits add support for handling phase shifts in `DataSet.hkl_to_observed()`. 

Although it is probably exceedingly unlikely to come across a use case with phases in an unmerged MTZ file, this support could be used in the future to write methods for changing `DataSet` objects to new spacegroups.

PR #11, opened 2020-10-28
new DataSet.compute_multiplicity method

kmdalton:
I was tired of dipping into `rs.utils` to compute multiplicities, so I made an `rs.DataSet.compute_multiplicity` method. This computes the multiplicity and adds a new `"EPSILON"` column to the data set.


JBGreisman (commented):
All looks good to me -- only comment is whether the multiplicity can ever be non-integer. Based on our discussion it seems like this should always be an integer, so I'm inclined to say it should return an `MTZInt` dtype instead of `MTZReal`. 

`MTZInt` automatically upcasts when mixed with a float-based `dtype` for arithmetic or `np.sqrt`, so I think it should work as expected.

JBGreisman (approved):
looks good to me

PR #12, opened 2020-10-29
MTZIntegerArray.to_numpy() should default to np.int32 where possible

JBGreisman:
The default behavior for a pandas `IntegerArray.to_numpy()` method is to return a numpy array of type `object`. This ensures support for NaN-values that can be represented in the `IntegerArray`, but not in a numpy array of `np.int32`. 

In order to minimize any surprise that may be caused by this default behavior, this PR changes the default behavior of `MTZIntegerArray.to_numpy()` to be to return a `np.int32` array if there are no NaN values in the array, but to keep the `object` return type if NaNs are present. This seems like a good compromise to maintain pandas-like behavior while improving default numpy conversions in many common cases.

### DIFF
--- a/reciprocalspaceship/dtypes/base.py
+++ b/reciprocalspaceship/dtypes/base.py
@@ -3,7 +3,7 @@ import numpy as np
 from pandas.core import nanops
 from pandas.core.indexers import check_array_indexer
 from pandas.core.construction import extract_array
-from pandas._libs import lib
+from pandas._libs import lib, missing as libmissing
 from pandas.api.extensions import (
     ExtensionDtype,
     ExtensionArray,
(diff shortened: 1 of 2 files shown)


reciprocalspaceship/dtypes/base.py
```diff
@@ -59,6 +59,48 @@ def _from_factorized(cls, values, original):
     def reshape(self, *args, **kwargs):
         return self._data.reshape(*args, **kwargs)
 
+    def to_numpy(self, dtype=None, copy=False, na_value=lib.no_default):
+        """
+        Convert to a NumPy Array.
+
+        If array does not contain any NaN values, will return a np.int32
+        ndarray. If array contains NaN values, will return a ndarray of 
+        np.float32 dtype.
```
kmdalton, line 68: I am confused, here it says it will return `np.float32` if `NaN`s are found, but below it looks like it returns `object`. Does a magical type conversion happen somewhere or is this just a typo? I am well outside my wheelhouse on this one, so i may be missing the point.

JBGreisman: This is a typo -- I forgot to update the docstring after our offline discussion.

PR #17, opened 2020-11-15
refactor inplace methods using two new decorators

kmdalton:
Decorators could help us insure that the behavior of `DataSet` methods are consistent with respect to indices and decrease the amount of boilerplate. I have created a provisional [branch](https://github.com/Hekstra-Lab/reciprocalspaceship/tree/decorator_refactor) to illustrate my point. 

In this branch, I have implemented two decorators, [`@inplacemethod`](https://github.com/Hekstra-Lab/reciprocalspaceship/blob/dacc8fb9b4a6ec4b5b2234915b816f9060d5e150/reciprocalspaceship/dataset.py#L24) and [`@with_range_index`](https://github.com/Hekstra-Lab/reciprocalspaceship/blob/dacc8fb9b4a6ec4b5b2234915b816f9060d5e150/reciprocalspaceship/dataset.py#L44). I used these two decorators to refactor all the methods in `rs.DataSet` which have an `inplace=Bool` parameter with the exception of the ones overloading `pandas` intrinsics. 

The `@inplacemethod` decorator just looks through the arguments at runtime and supplies `self.copy()` in place of `self` if `inplace==True`. This allowed me to remove all the explicit checks for `inplace` which had proliferated throughout the methods and replace them with a tidy decorator. 

The `@with_range_index` decorator supplies the same functionality as `@inplacemethod` inasmuch as it also checks for the value of `inplace` and does the right thing. Additionally it caches `self.index.names`, and converts `self.index` to a `RangeIndex` before passing it to the wrapped method. After the wrapped method returns, it applies whatever transformations are necessary to convert `self.index` back to the original format. I noticed we had several methods which called `self.reset_index` and later called `self.set_index` to try to restore the original index. However, these were implemented inconsistently and caused surprising behavior. Having a single centralized implementation for this sort of index manipulation should prevent these situations in the future. 

I'm going to go ahead and submit this as a pull request right now. However, I think we should have a lengthy discussion before this gets merged. It is a pretty involved refactor that touches a lot of methods. I think it is likely that we'll want to change some things before we go ahead. I also have no idea how badly this is going to break the documentation.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -17,6 +17,52 @@ from reciprocalspaceship.utils import (
     compute_structurefactor_multiplicity,
     bin_by_percentile
 )
+from functools import wraps
+from inspect import getcallargs
+
+
+def inplacemethod(f):
+    """ 
+    A decorator that applies the inplace argument. Base function must have a Bool param called "inplace".
+    The position of `inplace` doesn't matter.
+    """
+    @wraps(f)
+    def wrapped(self, *args, **kwargs):
+        locvars = getcallargs(f, self, *args, **kwargs)
+        if 'inplace' in locvars:
+            if locvars['inplace']:
+                return f(self, *args, **kwargs)
+            else:
+                return f(self.copy(), *args, **kwargs)
+        else:
+            print(locvars)
+            raise KeyError(f'"inplace" not found in local variables of @inplacemethod decorated function {f} '
+                             "Edit your method definition to include `inplace=Bool`. "
+                             )
+    return wrapped
+
+def with_range_index(f):
+    """ A decorator that presents the dataset with a range index and makes sure the method preserves the original index """
+    @wraps(f)
+    def wrapped(ds, *args, **kwargs):
+        locvars = getcallargs(f, ds, *args, **kwargs)
+        inplace = locvars.get('inplace', False)
+        if inplace:
+            ds = ds 
+        else:
+            ds = ds.copy()
+
+        names = ds.index.names
+        if names != [None]:
+            ds.reset_index(inplace=True)
+        ds = f(ds, *args, **kwargs)
+        if names == [None] and ds.index.names != [None]:
+            ds.reset_index(inplace=True)
+        elif names != [None] and ds.index.names == [None]:
+            ds.reset_index(inplace=True, drop=True)
+            ds.set_index(names, inplace=True)
+        return ds.__finalize__(ds)
+    return wrapped
 
 class DataSet(pd.DataFrame):
     """
@@ -476,7 +522,8 @@ class DataSet(pd.DataFrame):
         """
         keys = [ k for k in self if isinstance(self.dtypes[k], rs.M_IsymDtype) ]
         return keys
-    
+
+    @with_range_index
     def apply_symop(self, symop, inplace=False):
         """
         Apply symmetry operation to all reflections in DataSet object. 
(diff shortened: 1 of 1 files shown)


kmdalton:
One noteworthy shortcoming of this solution is that the logic to implement `inplace` functionality is duplicated in both of these decorators. If we refactor these decorators as classes, we could make `with_range_index` an extension of `inplacemethod`. This would force consistent handling of the `inplace` argument. However, these decorators don't require arguments at the moment, so I am hesitant to do this, because it would mean changing the tide `@inplacemethod` call signature to `@inplacemethod()`. 

Initially I was hoping to be able to "daisy chain" these decorators, ie
```python
@inplacemethod
@with_range_index
def hkl_to_asu(self, inplace=False):
    ...
```
However, it turned out to be too tricky to implement this. If we could figure out a way to do that, I think it'd be optimal.

reciprocalspaceship/dataset.py
```diff
@@ -17,6 +17,52 @@
...
+    @wraps(f)
+    def wrapped(self, *args, **kwargs):
+        locvars = getcallargs(f, self, *args, **kwargs)
+        if 'inplace' in locvars:
+            if locvars['inplace']:
+                return f(self, *args, **kwargs)
+            else:
+                return f(self.copy(), *args, **kwargs)
+        else:
+            print(locvars)
```
JBGreisman, line 38: Please remove print statement

JBGreisman (commented):
I definitely like this move towards decorators for avoiding reuse of snippets.  Along those lines, my main issue right now is the reuse of the `inplacemethod` code in `with_range_index`. I think it makes some sense to figure out how to daisy chain things together -- let's discuss how to make this happen.

JBGreisman:
Do you think `inplacemethod()` should just be called `inplace()`?

kmdalton:
I am worried that it might be close to impossible to make daisy chaining possible without moving to using classes instead of functions for decorators. Should we go ahead with this pull request after I remove the print statement and worry about daisy chaining later? Or do you want to deal with that first?

kmdalton:
I can rename `@inplacemethod`. For transparency -- I have exactly zero attachment to these names. I just knocked out this implementation to illustrate my point, and because it was the quickest way I could think of to fix the indexing inconsistencies I encountered.

kmdalton:
I just pushed a new version that uses daisy chained decorators. The key to getting it to behave to to switch to using the `inspect.Signature` for searching out the in `inplace` argument. This is passing tests on my machine, but it may need some more work. 

I added a new private function in the `rs.dataset` namespace which is `def _index_from_names(names, ds):`. I'm not sure where this should go. It might be best to make it a method of `DataSet`. Open for discussion.

reciprocalspaceship/dataset.py
```diff
@@ -17,6 +17,53 @@
...
+                return f(ds, *args, **kwargs)
+            else:
+                return f(ds.copy(), *args, **kwargs)
+        else:
+            raise KeyError(f'"inplace" not found in local variables of @inplacemethod decorated function {f} '
+                             "Edit your method definition to include `inplace=Bool`. "
+                             )
+    return wrapped
+
+def _index_from_names(names, ds):
```
JBGreisman, line 45: Fine as is, but this could also become a private method of `DataSet` if we would like. Then the call signature would be `def _index_from_names(self, names)`, which could be cleaner. I'm fine with it as is, too.

JBGreisman (approved):

PR #34, opened 2020-12-23
adding calculation for centric phase residual

apeck12:
Functions added for computing the phase residuals of centric reflections to their expected values.


kmdalton (changes requested):
This looks helpful, but would you mind modifying it to use `rs.utils.canonicalize_phases`? Then we can remove the `wraptopi` function definition. I think they are equivalent.

apeck12:
Thanks for pointing out `canonicalize_phases` in utils. The latest commit uses that function instead.

PR #45, opened 2021-01-28
Frenched wilson

kmdalton:
I redid the French Wilson scaling implementation. Notable differences from the previous version:
 - All posteriors are estimated by Gauss-Legendre quadrature
 - Structure factor means and standard deviations are estimated by quadrature. They were treated analytically before.
 - Both centric and acentric posteriors are refactored into a single function `_french_wilson_posterior_quad`
 - I'm using scipy.stats distributions to compute log probabilities 
 - I'm doing all calculations in log space using the scipy logsumexp function to compute summations

This PR fixes two important bugs
 - Structure factor posteriors were not computed correctly in the past. You cannot just take the square root of the posterior intensities. Instead we compute expectations for the moments. 
 - Previous versions didn't compute the mean intensity, `Sigma` correctly. Intensities should be corrected for their multiplicities before estimating `Sigma`. The resulting `Sigma` values should be re-inflated after by multiplying by the multiplicity.


reciprocalspaceship/algorithms/scale_merged_intensities.py
```diff
@@ -1,11 +1,13 @@
 import numpy as np
-from scipy.special import ndtr,log_ndtr
+import reciprocalspaceship as rs
 from reciprocalspaceship.utils import compute_structurefactor_multiplicity
```
JBGreisman, line 3: Let's remove this and use the newer `DataSet.compute_multiplicity()`

reciprocalspaceship/algorithms/scale_merged_intensities.py
```diff
@@ -245,49 +291,35 @@ def scale_merged_intensities(ds, intensity_key, sigma_key, output_columns=None,
...
-        outputI, outputSigI, outputF, outputSigF = output_columns
-    else:
-        columns = ["FW-I", "FW-SIGI", "FW-F", "FW-SIGF"]
-        outputI, outputSigI, outputF, outputSigF = columns
-        
+    if output_columns is None:
+        output_columns = ["FW-I", "FW-SIGI", "FW-F", "FW-SIGF"]
+    outputI, outputSigI, outputF, outputSigF = output_columns
+
+    multiplicity = compute_structurefactor_multiplicity(ds.get_hkls(), ds.spacegroup)
```
JBGreisman, line 298: `multiplicity = ds.compute_multiplicity()["EPSILON"].to_numpy()`

ought to work

JBGreisman (changes requested):
Looks good -- let's just make the small change to avoid the direct call to `rs.utils`.

JBGreisman (approved):

PR #48, opened 2021-02-02
Chevychase

kmdalton:
This pull request will 
 - improve quadrature integration for French Wilson scaling
 - add updated reference data from `pymc3`
 - add improved tests for posterior parameter estimation
 
The major point is that we now have a much larger gold standard test data set in the form of Markov chain Monte Carlo integration performed in `pymc3`. I've ensured that the output of our algorithm stays very close to the MCMC output. I've improved numerical stability and decreased memory requirement by using Gauss-Chebyshev quadrature with degree 100. It is possible we can lower this degree setting if there are performance concerns. Currently, all posterior parameters are within 2.5% of the MCMC values as compared to 6% for the cctbx implementation of the classical algorithm. 

I think this implementation constitutes a new state of the art. Two things we could potentially do to improve down the line:
 - optimize the integration window size in order to maximize consistency with MCMC
 - optimize the degree of the approximation
Both of these things are set quite arbitrarily right now. 

Here are some histograms of percent errors between `rs` and MCMC for the four posterior parameter estimates:

image

image

image

image


JBGreisman (approved):
Looks good to me. I'm not crazy about including the standalone plotting script `diagnostics.py`, but if you find it useful I think its fine. It's at least in the testing > data subdirectory specific to french-wilson so its context is fairly clear.

PR #49, opened 2021-02-10
Add license file to manifest

Anthchirp:
The license file is currently missing from the pypi distributions.


JBGreisman (approved):

JBGreisman:
Good catch -- thanks for the change! I'll get a new PyPI release up asap to resolve this.

PR #50, opened 2021-02-10
Add crystfel streams to `rs/io` module

marinegor:
This PR adds:
 - reading `CrystFEL's` stream files into `DataSet` format without any metadata
 - computing columns `ewald_offset`, `s1{x,y,z}` upon reading
 - accompanying tests
 - example of CrystFEL's stream to test on.

### DIFF
--- a/reciprocalspaceship/__init__.py
+++ b/reciprocalspaceship/__init__.py
@@ -1,11 +1,8 @@
 # Version number for reciprocalspaceship
 def getVersionNumber():
     import pkg_resources
-
     version = pkg_resources.require("reciprocalspaceship")[0].version
     return version
-
-
 __version__ = getVersionNumber()
 
 # Top-Level API
@@ -21,21 +18,21 @@ from reciprocalspaceship import algorithms
 # Add support for MTZ data types:
 # see http://www.ccp4.ac.uk/html/f2mtz.html
 from .dtypes import (
-    HKLIndexDtype,  # H
-    IntensityDtype,  # J
-    StructureFactorAmplitudeDtype,  # F
-    AnomalousDifferenceDtype,  # D
-    StandardDeviationDtype,  # Q
-    FriedelStructureFactorAmplitudeDtype,  # G
-    StandardDeviationFriedelSFDtype,  # L
-    FriedelIntensityDtype,  # K
-    StandardDeviationFriedelIDtype,  # M
-    NormalizedStructureFactorAmplitudeDtype,  # E
-    PhaseDtype,  # P
-    WeightDtype,  # W
-    HendricksonLattmanDtype,  # A
-    BatchDtype,  # B
-    M_IsymDtype,  # Y
-    MTZIntDtype,  # I
-    MTZRealDtype,  # R
+    HKLIndexDtype,                           # H
+    IntensityDtype,                          # J
+    StructureFactorAmplitudeDtype,           # F
+    AnomalousDifferenceDtype,                # D
+    StandardDeviationDtype,                  # Q
+    FriedelStructureFactorAmplitudeDtype,    # G
+    StandardDeviationFriedelSFDtype,         # L
+    FriedelIntensityDtype,                   # K
+    StandardDeviationFriedelIDtype,          # M
+    NormalizedStructureFactorAmplitudeDtype, # E
+    PhaseDtype,                              # P
+    WeightDtype,                             # W
+    HendricksonLattmanDtype,                 # A
+    BatchDtype,                              # B
+    M_IsymDtype,                             # Y
+    MTZIntDtype,                             # I
+    MTZRealDtype                             # R
 )
--- a/reciprocalspaceship/__init__.py
+++ b/reciprocalspaceship/__init__.py
@@ -1,14 +1,17 @@
 # Version number for reciprocalspaceship
 def getVersionNumber():
     import pkg_resources
+
     version = pkg_resources.require("reciprocalspaceship")[0].version
     return version
+
+
 __version__ = getVersionNumber()
 
 # Top-Level API
 from .dataset import DataSet
 from .dataseries import DataSeries
-from .io import read_mtz, read_precognition, read_csv
+from .io import read_mtz, read_precognition, read_csv, read_crystfel
 from .dtypes import summarize_mtz_dtypes
 from .concat import concat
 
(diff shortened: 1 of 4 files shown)


JBGreisman:
Overall, this looks mostly good to me. I'm gonna review and request a few minor revisions. 

High-level questions: 
- is a CrystFEL stream file only applicable to unmerged data? If so, we should explicitly set `DataSet.merged = False`
- as is, `DataSet.spacegroup` and `DataSet.cell` are left unset. Is the spacegroup or cell parameters accessible in the CrystFEL stream? It may be a good idea to parse them out if that's possible -- I'd imagine cell parameters may not be reasonable to get this way for a serial experiment, but perhaps they could be averaged in a meaningful way?
- It looks like `read_crystfel()` was given an optional `logfile=None` argument. I'm assuming this is a vestige from `read_precognition()`, but just wanted to check whether there are any additional arguments that this should take -- this could be a good place to add optional, user-specified spacegroup/cell parameters in case they aren't available in the stream itself.

JBGreisman:
I just noticed the test data is 98MB -- it looks like its a raw text with many "crystal" entries, that I'm assuming belong to a serial experiment. We try to keep this test data fairly minimal to what's needed to assess the core functionality. 

Do you mind reducing the data by only including something like 2 crystals? Whatever seems like a reasonable minimum number to test the parser.

reciprocalspaceship/__init__.py
```diff
@@ -18,21 +21,21 @@ def getVersionNumber():
 # Add support for MTZ data types:
 # see http://www.ccp4.ac.uk/html/f2mtz.html
 from .dtypes import (
```
JBGreisman, line 23: Do you mind reverting this file back to original and only adding `read_crystfel` to the top-level namespace? I know the file isn't quite "conventionally" formatted, but I find the aligned comments most clear

reciprocalspaceship/io/crystfel.py
```diff
@@ -0,0 +1,190 @@
+import pandas as pd
+from reciprocalspaceship import DataSet
+import numpy as np
+
+
+def _parse_stream(filename: str) -> dict:
```
JBGreisman, line 6: Nice parser -- clear to read. One revision -- please change the docstring to match that of `read_crystfel()`.

reciprocalspaceship/io/crystfel.py
```diff
@@ -0,0 +1,190 @@
...
+                # example line:
+                #    h    k    l          I   sigma(I)       peak background  fs/px  ss/px panel
+                #  -63   41    9     -41.31      57.45     195.00     170.86  731.0 1350.4 p0
+                crystal_peak_number += 1
+                h, k, l, I, sigmaI, _, _, _, _, _ = [i for i in line.split()]
+                h, k, l = map(float, [h, k, l])
+
+                # calculate ewald offset and s1
+                hkl = np.array([h, k, l])
+                # print(A)
```
JBGreisman, line 104: Please delete this line

reciprocalspaceship/io/crystfel.py
```diff
@@ -0,0 +1,190 @@
...
+
+            elif starts_crystal_peaks(line):
+                crystal_idx += 1
+                is_crystal = True
+                continue
+
+    return answ_crystals
+
+
+def read_crystfel(streamfile, logfile=None) -> DataSet:
```
JBGreisman, line 157: Please remove `logfile=None` from the call signature. It may be useful to add optional `spacegroup` and `cell`, but we can discuss elsewhere

reciprocalspaceship/io/crystfel.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    Initialize attributes and populate the DataSet object with data from a CrystFEL stream with indexed reflections. This is the output format used by CrystFEL software when processing still diffraction data.
+
+    Parameters
+    ----------
+    streamfile : str of file
+        name of a .stream file or a file object
+    """
```
JBGreisman, line 165: Please add:
```
Returns
--------
rs.DataSet
```

reciprocalspaceship/io/crystfel.py
```diff
@@ -0,0 +1,190 @@
...
+    # s1{x,y,z} -- R
+    # ewald_offset -- R
+    names = [
+        'H', 'K', 'L', 'I', 'sigmaI', 'BATCH', 's1x', 's1y', 's1z',
+        'ewald_offset'
+    ]
+    mtztypes = ["H", "H", "H", "J", "Q", "I", "R", "R", "R", "R"]
+    dataset = DataSet()
+    for (k, v), mtztype in zip(df.items(), mtztypes):
+        dataset[k] = v
```
JBGreisman, line 186: To my knowledge, these two lines can be simplified to:
`dataset[k] = v.astype(mtztype)`

reciprocalspaceship/io/crystfel.py
```diff
@@ -0,0 +1,190 @@
...
+                lambda_inv = 1 / eV2Angstrom(photon_energy)
+                s0 = np.array([0, 0, lambda_inv]).T
+
+            elif is_crystal:
+                # example line:
+                #    h    k    l          I   sigma(I)       peak background  fs/px  ss/px panel
+                #  -63   41    9     -41.31      57.45     195.00     170.86  731.0 1350.4 p0
+                crystal_peak_number += 1
+                h, k, l, I, sigmaI, _, _, _, _, _ = [i for i in line.split()]
+                h, k, l = map(float, [h, k, l])
```
JBGreisman, line 100: Should `h`, `k`, and `l` be cast to `int`? Or is it necessary to have floats for the subsequent matrix multply?

tests/io/test_crystfel.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertIsInstance(hewl["sigmaI"], rs.DataSeries)
+        self.assertIsInstance(hewl["ewald_offset"], rs.DataSeries)
+        self.assertIsNone(hewl.spacegroup)
+        self.assertIsNone(hewl.cell)
+
+        # chech values specific to the stream
+        self.assertTrue(len(hewl.index.unique()) < len(hewl.index))
+        self.assertTrue(
+            hewl['ewald_offset'].min() < 0 < hewl['ewald_offset'].max())
+        self.assertEqual(len(hewl.BATCH.unique()),
```
JBGreisman, line 46: If we reduce the number of crystals in the test data we'll need to modify this assertion

reciprocalspaceship/__init__.py
marinegor: Sure, do it. I just have `yapf` on by default, which messed already existing files.

JBGreisman (changes requested):
Thanks for the PR! On the whole it all looks great. All my comments were very minor. The only main things in my mind to change are:

1. Let's reduce the test file size
2. I think it would make sense to set the `spacegroup`, `cell`, and `merged` attributes, if at all possible

tests/io/test_crystfel.py
```diff
@@ -0,0 +1,49 @@
...
+def test_read_crystfel_mtz(IOtest_mtz):
+    """
+    rs.read_crystfel should raise ValueError when given file without
+    .stream suffix
+    """
+    with pytest.raises(ValueError):
+        rs.io.read_crystfel(IOtest_mtz)
+
+
+class TestPrecognition(unittest.TestCase):
```
JBGreisman, line 18: Rename to `TestCrystFEL` or something comparable

reciprocalspaceship/io/crystfel.py
marinegor: I didn't fully understand your remark regarding the docstring -- please check whether it's done as you intended.

reciprocalspaceship/io/crystfel.py
marinegor: Removed `logfile` and moved conversation about `spacegroup` and `cell` below

reciprocalspaceship/io/crystfel.py
marinegor: I copied this line from `read_precognition.py` -- I believed you had some reason to do that :)

reciprocalspaceship/io/crystfel.py
marinegor: You're right, `int` should be enough. `numpy` casts them to float during the calculation.

reciprocalspaceship/io/crystfel.py
JBGreisman: fair enough -- it certainly works as written, so we can just leave it. I cleaned up that section of `read_precognition()` last night, and we can always apply the same changes here.

marinegor:
Hi @JBGreisman , thanks for your detailed reply. I resolved most of your comments, here is the detailed reply:

> is a CrystFEL stream file only applicable to unmerged data? If so, we should explicitly set DataSet.merged = False
Only unmerged, right. The flag is now set.

> Is the spacegroup or cell parameters accessible in the CrystFEL stream? 
During `stream` stage of processing, the symmetry is considered to be unknown. The only things that `stream` contains are cell parameters, lattice type and centering (you can see the notation yourself in the stream header, search for 'Begin unit cell` line. If you know how to specify them in the `Dataset.spacegroup`, I'm open for your suggestions.

> It may be a good idea to parse them out if that's possible -- I'd imagine cell parameters may not be reasonable to get this way for a serial experiment, but perhaps they could be averaged in a meaningful way?
Sure, I added parsing of cell parameters from the header and added respective test for equality (`np.allclose`-ity, to be precise).

> Do you mind reducing the data by only including something like 2 crystals? Whatever seems like a reasonable minimum number to test the parser.
I choose 3 -- like odd numbers. My only with was for the stream to have crystals with both positive and negative `ewald_offset` for one of the tests -- I overdid this, for sure.

JBGreisman:
good catch with the `BatchDtype`

JBGreisman (approved):
Looks good to me. I'll merge this in shortly.

JBGreisman:
Thanks @marinegor!

PR #55, opened 2021-03-27
clearer error message in `to_gemmi`

kmdalton:
When `rs` runs into trouble converting a column to a `gemmi` compatible `dtype`, it is helpful to know the column name in addition to the offending dtype.


JBGreisman (approved):
This looks good to me. It is worth noting that this will only output the first offending column (if there are multiple), but I think this is an improvement over what is there now.

PR #58, opened 2021-04-02
Adding methods for interconverting wavelength and energy

kmdalton:
This PR will add two new methods to `rs.utils`, `ev2angstroms` and `angstroms2ev`.


wojdyr:
in gemmi the same is done by multiplying/dividing by `gemmi.hc`.
Checking the exact value of the constant, it's the same up to 9th significant digit: 
12398.41984332
12398.41973862
which I suppose is perfectly fine. I calculated the constant with Linux program units: 

    units -d15 'h * c / eV / angstrom'

JBGreisman (approved):
Looks good to me

Issue #59, opened 2021-05-10
Add a function to compute completeness


kmdalton:
I often find myself using external tools to compute completeness, but this is something we could easily implement within `rs`. I imagine there are other stats we might want to have baked in as well. I propose we add a `compute_completeness` function with an optional `bins` argument. I would be in favor of adding an `rs.stats` namespace for it to live in. 

How does this plan sit with you, @JBGreisman ?

JBGreisman:
yes -- I think this is a good idea. I think adding a `stats` subdirectory seems like a good place for such functions to live.

DHekstra:
I would mark this as low-priority, FWIW. 
I did do a simple implementation previously (https://github.com/Hekstra-Lab/double-wilson/blob/main/old/DoubleWIlsonExplorations.ipynb), but it's kind of slow.

kmdalton:
@DHekstra , I think I need this for processing the GFP data.

wojdyr:
Gemmi will have a function to calculate completeness at some point, but I don't know when.

JBGreisman:
Sounds good. Multiplicity and completeness would be useful stats to be able to compute, and could live in `rs.stats`. Computing those statistics using the `groupby` functionality in `pandas` will also make for a rather convenient way to summarize over resolution bins.

kmdalton:
I will go ahead and implement something to get us started.

kmdalton:
#60 introduces low level tools to `rs.utils` which can be used to address this issue. However, we still need to wrap the tools I added to become part of the top-level API or with a convenient function in the `rs.stats` module.

JBGreisman:
Here is a snippet that uses the stats function introduced in #60 and packages the results as a DataSet indexed by resolution bin. I'm just putting this here as a placeholder snippet for when I get around to implementing this:

```python
h, counts = rs.utils.compute_redundancy(mtz.get_hkls(), mtz.cell, mtz.spacegroup)
ds = rs.DataSet({"n": counts, "H": h[:, 0], "K": h[:, 1], "L": h[:, 2]}, spacegroup=mtz.spacegroup, cell=mtz.cell)
ds.set_index(["H", "K", "L"], inplace=True)
ds, labels = ds.assign_resolution_bins(10)
ds["observed"] = ds["n"] > 0
result = ds.groupby("bin")["observed"].agg(["sum", "count"])
result["completeness"] = result["sum"] / result["count"]
result.index = labels
[...]

PR #60, opened 2021-05-14
Adding a Stats Module and Multiplicity Calculator

kmdalton:
This branch attempts to address #59 . In this commit I have added a new `rs.stats` module. This contains a function for computing the multiplicity of a set of miller indices under a given a cell and spacegroup. One can cheaply get the completeness from the output of this function. I have also added tests, and two new methods to `rs.utils` in order to support stats. Since this is a fairly substantial change, I'm going to request that @JBGreisman do a thorough code review.


reciprocalspaceship/__init__.py
```diff
@@ -15,6 +15,9 @@ def getVersionNumber():
 # Import algorithms submodule
 from reciprocalspaceship import algorithms
 
+# Import stats submodule
+from reciprocalspaceship import stats
```
JBGreisman, line 19: Let's remove this from the top-level API -- as discussed offline, we can add a method to `DataSet` that wraps this call.

reciprocalspaceship/stats/completeness.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+import pandas as pd
+import reciprocalspaceship as rs
+
+def compute_multiplicity(hobs, cell, spacegroup, anomalous=False, dmin=None):
```
JBGreisman, line 5: Let's change this to `compute_redundancy()` to avoid confusion with `DataSet.compute_redundancy()`

reciprocalspaceship/stats/completeness.py
```diff
@@ -0,0 +1,56 @@
...
+    }, cell=cell, spacegroup=spacegroup).groupby(['H', 'K', 'L']).sum()
+
+    hall = rs.utils.generate_reciprocal_asu(cell, spacegroup, dmin, anomalous)
+
+    ASU = rs.DataSet({                     
+	'H' : hall[:,0],             
+	'K' : hall[:,1],             
+	'L' : hall[:,2],             
+	'Count' : np.zeros(len(hall)),
+    }, cell=cell, spacegroup=spacegroup).groupby(['H', 'K', 'L']).sum()
```
JBGreisman, line 53: Is this `groupby().sum()` call necessary? Shouldn't all the entries already be unique?

JBGreisman (approved):
On the whole, I think this looks good. Couple of general comments, not in any particular order:

- While I see that we can compute completeness with this `compute_redundancy()` function, should the file be named `completeness.py` at this point? 
- I think we should make a top-level `rs.stats.compute_redundancy(DataSet)` and perhaps `rs.stats.compute_completeness(DataSet)` that wraps this function and is user-facing. Do you think we should put this "numpy-backed" implementation in a private method? Or in `utils` where the rest of these types of functions live? 

Once we complete this PR, I can put together the above methods that wrap this and operate directly on unmerged `DataSet` objects for redundancy and unmerged or merged `DataSet` objects for completeness.

kmdalton:
- I believe I incorporated all your specific changes
 - I renamed the `completeness.py`  to `redundancy.py` which makes more sense as you noted
 
This function is maybe too useful to be totally hidden away as a private method. I am happy to move it to `utils` at your discretion. I think if we do move it to `utils`, it should not return a `DataSet`. Rather, it should return an `n x 3` array of miller indices and an `n` array of ints or floats. 
 
Let me know what you want to do. I'm happy to do whatever. For me the most useful things in this PR are the `generate_*` functions I added. I'm happy as long as those are not contentious.

JBGreisman (approved):
Looks good to me. My feeling is that we should move the `rs.stats.compute_redundancy()` method into `utils`, and then I can add functions to `stats/` that operate on DataSet objects by calling to that method.

kmdalton:
okay -- i moved `redundancy.py` into `utils/stats.py`. I cleaned up the imports and tests.

JBGreisman (approved):
Looks good -- I think this makes the most sense as you have it now, and can be easily wrapped to make user-facing, convenience methods that take `DataSet` objects.

Feel free to merge.

kmdalton:
Okay. I will merge this.

kmdalton:
Oops -- accidentally closed this. When I meant to merge. Nothing to see here, please move along...

PR #63, opened 2021-06-04
Fix sign of phase shift in DataSet.apply_symop

JBGreisman:
Based on visual inspection of maps, it appeared that `DataSet.apply_symop()` was applying phase shifts in the opposite direction of the specified symmetry operation. This PR changes the sign for how phase shifts are applied, and updates the relevant test.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -1000,7 +1000,7 @@ class DataSet(pd.DataFrame):
         for k in dataset.get_phase_keys():
             dataset[k] = phi_coeff[inverse] * (dataset[k] + phi_shift[inverse])
         dataset.canonicalize_phases(inplace=True)
-
+        
         # GH#3: if PARTIAL column exists, use it to construct M/ISYM
         if "PARTIAL" in dataset.columns:
             m_isym = isym[inverse] + 256*dataset["PARTIAL"].to_numpy()
@@ -1085,6 +1085,10 @@ class DataSet(pd.DataFrame):
         # Apply phase shift
         for k in self.get_phase_keys():
             self[k] = phi_coeff[inverse] * (self[k] + phi_shift[inverse])
+        for k in self.get_complex_keys():
+            self[k] *= np.exp(1j*np.deg2rad(phi_shift[inverse]))
+            friedel_mask = phi_coeff[inverse] != 1
+            self.loc[friedel_mask, k] = np.conjugate(self.loc[friedel_mask, k])
         self.canonicalize_phases(inplace=True)
         
         return self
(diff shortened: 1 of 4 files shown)


JBGreisman:
To do:
 - It would be useful to make a "gold standard" pair of F/phi with a known real-space translation between the corresponding maps so that we can write a cleaner test.

JBGreisman:
My most recent push added a test that uses fmodel data to compute a real-space map, shifts the map based on a desire translation using `np.roll`, back computes corresponding complex structure factors. These are then compared to the results of `DataSet.apply_symop()`

reciprocalspaceship/dataset.py
```diff
@@ -597,8 +597,7 @@ def apply_symop(self, symop, inplace=False):
 
         # Shift phases according to symop
         for key in dataset.get_phase_keys():
-            dataset[key] += np.rad2deg(phase_shifts)
-            dataset[key] *= phic
+            dataset[key] = phic * (dataset[key] - np.rad2deg(phase_shifts))
             dataset[key] = utils.canonicalize_phases(dataset[key], deg=True)
         for key in dataset.get_complex_keys():
             dataset[key] *= np.exp(1j*phase_shifts)
```
kmdalton, line 603: Don't you need to change the sign of this complex rotation too?

kmdalton:
As per our conversation offline, let's go ahead and merge this and then do a refactor.

kmdalton (changes requested):
I am not clear why you did not change the sign of the phase shift applied to complex structure factors. Not sure if this is in error. If tests are passing, I am concerned that complex phase shifts are maybe not adequately tested. 

Can you either change the sign of complex phase shifts or tell me that there is no need for some reason that I am missing?

JBGreisman:
I am still working on this for the complex structure factors. This test may not be working as well as I had hoped, but I am not yet ready to come to that conclusion

JBGreisman:
This PR now fixes `DataSet.apply_symop()` for use with both |F|+phi and complex SFs. I updated the test to use a P1 dataset avoid an interdependency caused by the use of `apply_symop()` in `DataSet.to_reciprocalgrid()`. 

P.S. -- I updated my emacs recently to auto-apply `black` formatting, which is why there are additional changes to the `test_dataset_symops.py` file

reciprocalspaceship/dataset.py
JBGreisman: Fixed in latest update -- the phase shifts have now been flipped for complex structure factors as well

tests/test_dataset.py
```diff
@@ -495,14 +495,14 @@ def test_apply_symop_roundtrip(mtz_by_spacegroup):
     dataset = rs.read_mtz(mtz_by_spacegroup)
     for op in dataset.spacegroup.operations():
         applied = dataset.apply_symop(op)
-        back = applied.hkl_to_asu()
```
kmdalton, line 498: We should try to resurrect this test when we do the refactor, but this is okay with me for now.

kmdalton (approved):
This looks good to me.

PR #67, opened 2021-06-16
mostly vectorized phase restriction calculation

kmdalton:
I made a version of the phase shift calculation that has pretty much everything outside the `for` loop. It is about 3x faster in the test. Hard to say how it would perform with higher resolution data set. This was mostly an exercise to see what would be required to speed this up. I am not going to say whether we should actually merge this or not.


JBGreisman:
This seems better than what was there, so it looks good to me

JBGreisman (approved):

PR #68, opened 2021-06-16
Fix behavior of rs.utils.generate_reciprocal_asu() with anomalous flag

JBGreisman:
This PR fixes the behavior of `rs.utils.generate_reciprocal_asu()` with `anomalous=True` to avoid copying centric reflections to the Friedel-minus ASU. This makes the function's handling of centrics consistent with other methods (such as `DataSet.stack_anomalous()`)

I also changed the implementation to use `rs.utils.in_asu()` instead of `rs.utils.hkl_to_asu()`, which speeds up the code ~2x.


kmdalton:
Nice work!

PR #72, opened 2021-07-21
Modify Crystfel Stream Parser to Return Pixel Coordinates

kmdalton:
Modify the crystfel stream parser to return columns for the pixel position of spots. These are called "XDET" and "YDET" following the DIALS (and maybe others?) convention.


JBGreisman (approved):
looks good to me

PR #73, opened 2021-07-22
Compute the angular Ewald offset when loading streams

kmdalton:
This PR adds the angular Ewald offset to the metadata computed when loading stream files.


JBGreisman (approved):
looks good to me

JBGreisman:
In addition to computing angular Ewald offsets, this PR also speeds up the loading of unmerged datasets

PR #74, opened 2021-07-28
add support for generators to rs.concat

kmdalton:
`rs.concat` does not support generators whereas `pd.concat` does. The reason `rs` does not support them, is that the control flow that checks whether the arguments are isomorphous specifically [indexes the first element in the passed sequence](https://github.com/Hekstra-Lab/reciprocalspaceship/blob/3e66ad04ffeb69eccb277d1afb2ab9c0a4a9defa/reciprocalspaceship/concat.py#L32). This only works for subscriptable objects. Instead, implementing this control flow using `itertools.tee` allows us to bypass the requirement that the passed sequence be subscriptable.

### DIFF
--- a/reciprocalspaceship/concat.py
+++ b/reciprocalspaceship/concat.py
@@ -1,4 +1,5 @@
 import reciprocalspaceship as rs
+from itertools import tee
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


JBGreisman (approved):
Looks good to me

PR #75, opened 2021-07-28
added an angle_between function to rs.utils

kmdalton:
This pr adds a function to `rs.utils` for computing the angle between vectors in a numerically stable way. It also adds a test for said function. I also updated the `crystfel` stream file parser to use this new angle calculation in computing the angular ewald offset.


JBGreisman (commented):
Looks good to me -- once the conflicts are resolved I'll merge it in

JBGreisman:
I resolved the conflict (it was a simple thing), and will merge after CI finishes

PR #76, opened 2021-08-09
Added convenience function to remove systematic absences

kmdalton:
I added a method to `rs.DataSet` which just drops systematically absent rows. This saves the user having to do something like
```python
ds = ds[~ds.label_absences().ABSENT]
```
which can now by done by
```python
ds = ds.remove_absences()
```
. This method supports the `inplace` argument, and by using `pd.DataFrame.drop` and operating at the `ds.index` level, I think it avoids using extra memory unnecessarily. This will be helpful for large, unmerged datasets. The recipe was a tad tricky to work out, so I figured we should just put it in a method. (Admittedly this part is a premature optimization, but I do think the method makes code cleaner at the user level). 

@JBGreisman , let me know what you think

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -678,7 +678,10 @@ class DataSet(pd.DataFrame):
         -------
         DataSet
         """
-        return self[~self.label_absences().ABSENT]
+        mask = is_absent(self.get_hkls(), self.spacegroup)
+        idx = self.index[mask]
+        self.drop(index=idx, inplace=True)
+        return self
 
     @inplace
     def infer_mtz_dtypes(self, inplace=False, index=True):
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -665,6 +665,22 @@ class DataSet(pd.DataFrame):
         return self
 
     @inplace
+    def remove_absences(self, inplace=False):
+        """
+        Remove systematically absent reflections in DataSet. 
+
+        Parameters
+        ----------
+        inplace : bool
+            Whether to add the column in place or to return a copy
+
+        Returns
+        -------
+        DataSet
+        """
+        return self[~self.label_absences().ABSENT]
+
+    @inplace
     def infer_mtz_dtypes(self, inplace=False, index=True):
         """
         Infers MTZ dtypes from column names and underlying data. This 
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -225,7 +225,36 @@ def test_label_absences(data_fmodel, inplace, no_sg):
         assert "ABSENT" in result
         assert result["ABSENT"].dtype.name == "bool"
 
+@pytest.mark.parametrize("inplace", [True, False])
+@pytest.mark.parametrize("hkl_index", [True, False])
+def test_remove_absences(inplace, hkl_index):
+    """Test DataSet.remove_absences()"""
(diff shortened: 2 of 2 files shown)


tests/test_dataset.py
```diff
@@ -225,7 +225,36 @@ def test_label_absences(data_fmodel, inplace, no_sg):
...
+        ds.set_index(['H', 'K', 'L'], inplace=True)
+
+    ds_test = ds.remove_absences(inplace=inplace)
+    ds_true = ds[~ds.label_absences().ABSENT]
+
+    assert len(ds_test) == len(Hall) - absent.sum()
+    assert np.array_equal(ds_test.get_hkls(), ds_true.get_hkls())
+
+    if inplace:
+        assert id(ds_test) == id(ds)
```
JBGreisman, line 255: let's also add a clause that tests that `inplace=False` has `id(ds_test) != id(ds)`.

tests/test_dataset.py
```diff
@@ -225,7 +225,36 @@ def test_label_absences(data_fmodel, inplace, no_sg):
         assert "ABSENT" in result
         assert result["ABSENT"].dtype.name == "bool"
 
+@pytest.mark.parametrize("inplace", [True, False])
+@pytest.mark.parametrize("hkl_index", [True, False])
+def test_remove_absences(inplace, hkl_index):
```
JBGreisman, line 230: This test certainly works, but I think a more thorough test can be done using the `data_fmodel` fixture that tests all sohncke spacegroups with P1/not P1. I don't think its necessary here though given that `label_absences()` is tested in that manner.

reciprocalspaceship/dataset.py
```diff
@@ -664,6 +664,25 @@ def label_absences(self, inplace=False):
...
+        inplace : bool
+            Whether to add the column in place or to return a copy
+
+        Returns
+        -------
+        DataSet
+        """
+        mask = is_absent(self.get_hkls(), self.spacegroup)
+        idx = self.index[mask]
+        self.drop(index=idx, inplace=True)
```
JBGreisman, line 683: Is this intended?  Doesn't this function always operate as if `inplace=True`?

tests/test_dataset.py
kmdalton: done.

JBGreisman (changes requested):
Looks good -- only qualm is that `inplace=False` should return a DataSet with a different location in memory, so we should add that to the tests. I haven't written code for the `inplace` decorator in a while, so I'm not sure if it works as intended here without that test.

tests/test_dataset.py
JBGreisman: I'm fine with this one as is

reciprocalspaceship/dataset.py
kmdalton: i don't know what the antecedent of this is in your second sentence. if you are referring to `pd.DataFrame.drop` the answer is that it does take an `inplace` argument which defaults to `False`. if you are referring to `rs.DataSet.remove_absences`, the answer is that it only operates `inplace` if `inplace=True`. Otherwise the `@inplace` decorator creates a copy before calling the method. 

Does that answer your question?

JBGreisman: yeah -- i had forgotten whether the `@inplace` decorator had already handled making the copy

JBGreisman (approved):
With the added clause to the test, this looks good to me.

PR #86, opened 2021-08-25
Standardize code style using black

JBGreisman:
This is an initial look at using `black` to standardize the code style in `rs`. This initial set of commits just runs `black` on the existing code base.

To do:
- [x] automate running `black` on any PRs/pushes to standardize any new contributions.
- [x] document code style/best practices in developer section

PR #90, opened 2021-09-08
Fix reversion in dataset.py (#87)

JBGreisman:
When applying `black` to the codebase (#86), it appears that I reverted `DataSet.compute_dHKL()` to not use the changes made in #87. This PR corrects that reversion. 

After merging in this PR, `DataSet.compute_dHKL()` will use `gemmi` for computing resolution of reflections.


kmdalton (approved):
looks good.

Issue #18, opened 2020-11-17
DataSet.cell and DataSet.spacegroup attributes do not do type-checking


JBGreisman:
Currently, the `DataSet.cell` and `DataSet.spacegroup` attributes do not do any type checking. They are intended to be set to `gemmi.UnitCell` and `gemmi.SpaceGroup` objects, but they can currently be set to anything:
```python
mtz.spacegroup = [1, 2, 3]
print(mtz.spacegroup) # prints [1, 2, 3]
```
It would make sense to add some type checking here in their respective setter methods. This can also be made to broaden the API so that if spacegroups are set to be a string or int, the `gemmi.SpaceGroup` constructor is called, and same thing for `gemmi.UnitCell` if a list/tuple of 6 values is passed.

PR #94, opened 2021-09-13
Add decorators for gemmi argument checking

### LINKED ISSUE #92: [Feature Request] gemmification decorators
Decorators that scan the arguments of a function and automatically convert `cell` and `spacegroup` arguments to their proper `gemmi` types would be useful in writing standalone functions. We have 3 decorators

 - a general purpose `gemmify` decorator. @gemmify  scans for keyword arguments named sg, spacegroup , space_group, cell, unitcell, unit_cell, and converts them accordingly
 - a `spacegroupify` decorator which has `*args` for each argument name to be converted

```python
@spacegroupify("parent_sg", "child_sg")
def a_crazy_efx_function(parent_sg, child_sg, data):
    ....
```
 - a `cellify` decoraor (in the same vein of `spacegroupify`)

JBGreisman:
Fixes #92.

It is common in `rs` to write functions that take `gemmi.SpaceGroup` or `gemmi.UnitCell` instances as arguments. In order to make these methods more flexible, it is often useful to add error checking code so that if an `int` is provided instead of a `gemmi.SpaceGroup`, the method generates the equivalent `gemmi` object instance. This was done for the `DataSet.cell` and `DataSet.spacegroup` setter methods to address #18.

This PR generalizes this idea using `@spacegroupify` and `@cellify` decorators to avoid needing to write boiler plate argument checking code. This allows one to write a function as if it is guaranteed that the input argument is a `gemmi` object, while still being able to input that argument flexibly. For example:

```python
from reciprocalspaceship.decorators import spacegroupify

@spacegroupify
def get_xhm(spacegroup=None):
    return spacegroup.xhm()

import gemmi
print(get_xhm(gemmi.SpaceGroup("1"))) # prints 'P 1'
print(get_xhm(1))                     # prints 'P 1'
print(get_xhm("P1"))                  # prints 'P 1'
```

### DIFF
--- a/reciprocalspaceship/utils/structurefactors.py
+++ b/reciprocalspaceship/utils/structurefactors.py
@@ -2,6 +2,7 @@ import numpy as np
 from gemmi import GroupOps, SpaceGroup
 
 import reciprocalspaceship as rs
+from reciprocalspaceship.decorators import spacegroupify
 
 
 def to_structurefactor(sfamps, phases):
@@ -54,6 +55,7 @@ def from_structurefactor(sfs):
     return sf, phase
 
 
+@spacegroupify
 def compute_structurefactor_multiplicity(H, sg, include_centering=True):
     """
     Compute the multiplicity of each reflection in ``H``.
@@ -62,7 +64,7 @@ def compute_structurefactor_multiplicity(H, sg, include_centering=True):
     ----------
     H : array
         n x 3 array of Miller indices
-    spacegroup : gemmi.SpaceGroup, gemmi.GroupOps
+    spacegroup : str, int, gemmi.SpaceGroup
         The space group to identify the asymmetric unit
     include_centering : bool
         Whether or not to include the multiplicity inherent in the lattice centering.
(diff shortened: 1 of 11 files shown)


JBGreisman:
To do:
- [ ] Add `@gemmify` decorator as a convenience method for applying both `cellify(spacegroupify(function))`
- [ ] Use decorators in relevant parts of the `rs` codebase

reciprocalspaceship/utils/structurefactors.py
```diff
@@ -75,16 +77,7 @@ def compute_structurefactor_multiplicity(H, sg, include_centering=True):
...
-    if isinstance(sg, SpaceGroup):
-        group_ops = sg.operations()
-    elif isinstance(sg, GroupOps):
-        group_ops = sg
-    else:
-        raise ValueError(
-            f"gemmi.SpaceGroup or gemmi.GroupOps expected for parameter sg. "
-            f"Received object of type: ({type(sg)}) instead."
-        )
-
```
JBGreisman, line 87: Note: This change removes support for invoking this function directly with a `GroupOps` object. We do not actually use this anywhere in our codebase, so it seems like the added flexibility of `@spacegroupify` is preferable. This is also true for `is_absent()` and `is_centric()`.

Issue #97, opened 2021-10-02
stack/unstack_anomalous roundtrip does not work


kmdalton:
Calling `DataSet.unstack_anomalous` followed by `DataSet.stack_anomalous` does not always work in `rs` version `0.9.15`.  The following code verifies that this fails sometimes and succeeds others. 

```python
import gemmi
import reciprocalspaceship as rs
import numpy as np



cell = gemmi.UnitCell(10., 20., 30., 90., 90., 90.)
sg = gemmi.SpaceGroup(19)
dmin = 2.

h,k,l = rs.utils.generate_reciprocal_asu(cell, sg, dmin, anomalous=True).T
n = len(h)

ds = rs.DataSet({
        'H' : h,
        'K' : k,
        'L' : l,
        'F' : np.random.random(n),
        'loc' : np.random.random(n),
        'scale' : np.random.random(n),
    }, 
    spacegroup=sg, 
    cell=cell, 
    merged=True,
).infer_mtz_dtypes().set_index(['H', 'K', 'L'])

assert all(ds.keys() == ['F', 'loc', 'scale'])

unstacked = ds.unstack_anomalous()
print(unstacked.keys())
unstacked.stack_anomalous()
```

I find that the order of columns in `unstacked` is not always the same, despite consistent column ordering in `ds`. When the column order is `Index(['F(+)', 'loc(+)', 'scale(+)', 'F(-)', 'loc(-)', 'scale(-)'], dtype='object')`, the script succeeds. When the column order is `Index(['F(+)', 'loc(+)', 'scale(+)', 'scale(-)', 'loc(-)', 'F(-)'], dtype='object')`, it fails with the following traceback:
```
Traceback (most recent call last):                                                                                                                             
  File "stack_bug.py", line 31, in <module>                                                                                                                    
    unstacked.stack_anomalous()                                                                                                                                
  File ".../anaconda/envs/careless/lib/python3.8/site-packages/reciprocalspaceship/dataset.py", line 911, in stack_anomalous                    
    raise ValueError(                                                                                                                                          
ValueError: Corresponding labels in ['F(+)', 'loc(+)', 'scale(+)'] and ['scale(-)', 'loc(-)', 'F(-)'] are not the same dtype: FriedelSFAmplitude and MTZReal   
```
[...]

PR #99, opened 2021-10-05
Ensure consistent column ordering in `DataSet.unstack_anomalous()`

JBGreisman:
`DataSet.unstack_anomalous()` would sometimes produce new two-column anomalous `DataSets` in which the newly created anomalous columns were in different orders. This would occur because internally the function used a `set` to keep track of the new columns. Since the plus-Friedel columns were suffixed with `(+)` and the minus-Friedel with `(-)`, this could produce different orders to the columns in the `set`. 

This bug made it so that round-trip operations of `DataSet.unstack_anomalous()` followed by `DataSet.stack_anomalous()` sometimes produced a `ValueError` (#97 ), depending on the Python version being used. 

This PR updates `DataSet.unstack_anomalous()` to internally use a `dict` for keeping track of the relevant columns. As of Python 3.7+ this preserves the order of input keys.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -951,11 +951,11 @@ class DataSet(pd.DataFrame):
             dataset_minus, how="outer", on=["H", "K", "L"], suffixes=suffixes
         )
 
-        # Handle centric reflections
+        # Update centric reflections to use "aimless"-style formatting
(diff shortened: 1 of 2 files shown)


kmdalton:
I checked this branch and it appears to fix the issues i was having with `careless` output. :+1:

dennisbrookner:
Noted re: locations of conversations, I'll make a new Issue and link to here. Glad this seems to make sense!

PR #101, opened 2021-10-08
Add python 3.10 to test suite for github actions

JBGreisman:
Looking ahead, it makes sense to include this in our test suite to preemptively catch any issues


JBGreisman:
This PR now seems good to go to me. The only problem is that `pytorch`, which is a requirement in our `[dev]` build does not support `python3.10` yet. As such, a call to `pip install -e .[dev]` would currently fail in python3.10. 

I'm not sure if the above failure mode should keep us from updating to support python 3.10 though.

kmdalton (approved):
I don't have any serious issues with this PR. I don't think it is very likely the `dev` build issue will affect users. If it does, hopefully they'll submit an issue. 

However, I do wonder whether you can replace `python -m pip ....` with `pip ...` in these `yaml` files. I'm guessing they are written this way for a reason, but if we could `pip ...` it'd be slightly cleaner.

PR #111, opened 2021-11-08
`DataSet.write_mtz()` should not move unmerged reflections to reciprocal ASU

### LINKED ISSUE #110: `DataSet.write_mtz()` with unmerged DataSet moves reflections to ASU
If a DataSet object contains unmerged reflections, a call to `DataSet.write_mtz()` currently has the side effect of calling `DataSet.hkl_to_asu(inplace=True)` on the object:

```
In [1]: ds = rs.read_mtz("data_unmerged.mtz")

In [2]: ds["BATCH"].head()
Out[2]: 
H    K    L 
19   -11  2    1424
3    18   -1    328
-17  21   17    576
21   26   -7     70
-9   17   11    798
Name: BATCH, dtype: Batch

In [3]: ds.write_mtz("/dev/null")

In [4]: ds["BATCH"].head()
Out[4]: 
H   K   L 
19  11  2    1424
18  3   1     328
21  17  17    576
26  21  7      70
17  9   11    798
Name: BATCH, dtype: Batch
```

I consider this to be a bug because writing data to a file should not change the underlying object.

JBGreisman:
This PR fixes #110 to ensure that unmerged reflections are not moved to the reciprocal ASU after a call to `DataSet.write_mtz()`.

### DIFF
--- a/reciprocalspaceship/io/mtz.py
+++ b/reciprocalspaceship/io/mtz.py
@@ -4,6 +4,7 @@ import pandas as pd
 
 from reciprocalspaceship import DataSet
 from reciprocalspaceship.dtypes.base import MTZDtype
+from reciprocalspaceship.utils import in_asu
 
 
 def from_gemmi(gemmi_mtz):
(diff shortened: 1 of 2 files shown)


JBGreisman:
The most recent commit also extends this PR to support invoking `DataSet.write_mtz()` or `DataSet.to_gemmi()` with unmerged data that is already in the reciprocal ASU. Previously, such data would have been re-mapped back to the observed indices upon a call to `write_mtz()` or `to_gemmi()`

kmdalton (approved):

PR #113, opened 2021-11-08
vectorize utils.in_asu

kmdalton:
For some reason we never vectorized `utils.in_asu`. This fixes that and removes some unused variables. EOM.


JBGreisman (approved):
Looks good to me -- will merge once CI finishes

PR #128, opened 2022-01-23
Raise ValueError when stack_anomalous() will result in duplicate column names

dennisbrookner:
This was pretty straightforward. Now, if you attempt to call `DataSet.stack_anomalous()` on a dataset containing columns such as `I`, `I(+)`, and `I(-)`, you are greeted with the error message:
```
ValueError: Stacking anomalous data will result in duplicate column names. Rename or drop column 'I' and try again.
```
I also added `test_stack_anomalous_duplicates()` to `test_dataset_anomalous.py` to verify this behavior.

### Minor reordering
In the previous version of `DataSet.stack_anomalous()`, the line:
```python
new_labels = [l.rstrip(suffixes[0]) for l in plus_labels]
```
comes after the following:
```python
centrics = self.label_centrics()["CENTRIC"]
dataset_plus = self.drop(columns=list(minus_labels))
dataset_minus = self.loc[~centrics].drop(columns=list(plus_labels))
dataset_minus.apply_symop(gemmi.Op("-x,-y,-z"), inplace=True)
```
I was concerned that if I didn't check for bad labels until these four lines had already been run, that when the function exits with an error, the `DataSet` might still get modified without the user realizing? I'm not entirely sure if that's how things work, but in any event, I reordered things. Now, the labels are checked for potential duplicates *before* the `DataSet` is modified.

#### A random and annoying thing that I changed
Spyder astutely pointed out to me that several tests in `test_dataset_anomalous.py` contained things like:
```python
with pytest.raises(ValueError):
    result = data_unmerged.unstack_anomalous()
```
but never actually access the `result`. I changed these instances to
```python
with pytest.raises(ValueError):
    data_unmerged.unstack_anomalous()
```
etc. This is totally unrelated to this PR, so I'm happy to take this out if making this change here is annoying for code/version organization purposes.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -897,10 +897,10 @@ class DataSet(pd.DataFrame):
         for lab in new_labels:
             if lab in self.columns:
                 raise ValueError(
-                    f"Stacking anomalous data will result in duplicate column " 
+                    f"Stacking anomalous data will result in duplicate column "
                     f"names. Rename or drop column '{lab}' and try again."
                 )
-        
+
         # Map Friedel reflections to +/- ASU
         centrics = self.label_centrics()["CENTRIC"]
         dataset_plus = self.drop(columns=list(minus_labels))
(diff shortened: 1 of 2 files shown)


JBGreisman (approved):
This PR looks good to me. Thanks!

Issue #135, opened 2022-03-09
`rs.read_precognition()`: Update docstring and set `spacegroup` after parsing log


JBGreisman:
It came up today that `rs.read_precognition()` reads .ii files. This isn't _explicitly_ stated in the docstring. It should be made more clear that the io function supports both file types.

dennisbrookner:
Related to this issue, there's a bug in the logfile parsing of `rs.read_precognition()` where the spacegroup is parsed from the log file, but never actually assigned to the `DataSet`. I'll fix both of these in the same PR

dennisbrookner:
Following additional discussion offline, this PR will
- [ ] Clarify in docs that `.hkl` and `.ii` files are allowed
- [ ] Fix bug in which space group is not parsed from log files
- [ ] Clarify in docs that user-supplied space group and/or cell will take precedence over those found in log file
- [ ] Add a warning for when user supplies both a space group and/or cell and a log file

JBGreisman:
Awesome -- thanks! Two small clarifications:
1. I think the spacegroup is parsed correctly, it is just not set as a `DataSet` attribute. 
2. I think the warning should only be raised if the user supplies all 3 arguments (sg, cell, and log). I could see a valid use to providing just an sg or a cell if you only want to manually overwrite one attribute. I think the docstring should specify the precedence, but a warning should only be raised when the logfile parsing will not be used.

PR #136, opened 2022-03-13
Fixed `read_precognition()` as per #135 and updated tests

dennisbrookner:
Accomplished all of the changes mentioned in #135 . Two outstanding thoughts:

 - Though I did clarify in the docs that `.ii` files are allowed, I didn't want to change the name of the `hklfile` argument for fear of breaking backwards compatibility. That said, I know this function hasn't really seen a lot of daylight, so I'm happy to be overruled there.
 - I added a `ValueError` for when a `cell` argument is provided but doesn't have length 6. Previously, the function checked for this length, but then just silently ignored an argument with the wrong length. I didn't add a test for this behavior, thinking that was overkill, but I could add that if desired.

Also, as best I can tell, this is the first "warning" in `reciprocalspaceship`. If some mechanism other than the built-in `warnings` module is preferable, let me know.

### DIFF
--- a/reciprocalspaceship/io/precognition.py
+++ b/reciprocalspaceship/io/precognition.py
@@ -43,7 +43,7 @@ def read_precognition(hklfile, spacegroup=None, cell=None, logfile=None):
             F.rename(columns={"F(+)": "F", "SigF(+)": "SigF"}, inplace=True)
             mtztypes = ["H", "H", "H", "F", "Q"]
 
-    # Read data from II file    
+    # Read data from II file
     elif hklfile.endswith(".ii"):
         usecols = range(10)
         F = pd.read_csv(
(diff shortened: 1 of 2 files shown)


kmdalton:
This function should use the cellify and spacegroupify decorators. Type checking the cell and spacegroup should happen exclusively inside those.

kmdalton:
Jack will reply later, but imo type checks should live in the cellify decorator which is called by the setter.

dennisbrookner:
> Jack will reply later, but imo type checks should live in the cellify decorator which is called by the setter.

Yeah, Jack slacked me about this as well. Makes sense, I'll take that out.

JBGreisman (approved):
This looks good to me -- thanks for the PR!

JBGreisman:
As mentioned above, the setter methods for `DataSet.cell` and `DataSet.spacegroup` implicitly handle all the type checking and coercion to `gemmi` objects. These use the `@spacegroupify` and `@cellify` decorators to ensure that all the relevant type-checking is handled in one place. This change was made in #94.

As such, this looks good -- when the `DataSet` attributes are set any necessary checks/conversion to `gemmi` objects will be done.

PR #149, opened 2022-05-19
Remove `DataSet.append()` due to upcoming deprecation in Pandas

### LINKED ISSUE #146: Remove support for `DataSet.append()` due to pandas deprecation
`DataFrame.append()` and `Series.append()` were deprecated in pandas v1.4. We should remove the overloaded functions from `reciprocalspaceship` to avoid future compatibility issues. `rs.concat()` already has the required functionality to replace the methods.

JBGreisman:
This PR fixes #146 to remove the overloaded function, `DataSet.append()`, and any uses of it.


kmdalton (approved):
This looks good to me. I was not able to find any other places in the code base with offending `append` statements. To the best of my knowledge, this is complete. The straightforward substitution with `rs.concat` seems the most natural route too. Sorry for the late reply!

PR #151, opened 2022-05-22
pushing intensity.py and associated test

dennisbrookner:
I think this should do it, but let me know if there are any functionality and/or style things I'm missing.

### DIFF
--- a/reciprocalspaceship/algorithms/__init__.py
+++ b/reciprocalspaceship/algorithms/__init__.py
@@ -1,3 +1,6 @@
+from reciprocalspaceship.algorithms.intensity import (
+    compute_intensity_from_structurefactor,
+)
 from reciprocalspaceship.algorithms.merge import merge
 from reciprocalspaceship.algorithms.scale_merged_intensities import (
     scale_merged_intensities,
--- a/reciprocalspaceship/algorithms/intensity.py
+++ b/reciprocalspaceship/algorithms/intensity.py
@@ -1,0 +1,35 @@
+import numpy as np
+
+import reciprocalspaceship as rs
+
+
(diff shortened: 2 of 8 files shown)


dennisbrookner:
Sorry for making a huge mess of this! I'll fix it tomorrow

JBGreisman:
sounds good -- ping me when you're to have this looked at

dennisbrookner:
This is now functional! @JBGreisman let me know if I'm handling input types correctly; it seemed consistent with other similar functions to accept lists, `numpy` arrays, and `rs.DataSeries`.

JBGreisman (changes requested):
I think the API for this should match the other functions in `rs.algorithms` (see `merge()` or `scale_merged_intensities()`):

`compute_intensity_from_structurefactor(dataset, f_key="F", sigf_key="SigF")`

I think this will have the advantage of streamlining your error handling -- you just need to confirm that the selected columns have the correct MTZDtype. You can implement the function as it is, and return a new `DataSet` with desired columns. All this has the advantage that in the future we can implement this to be "smarter" about the distributions if we want --  that may require multiplicities, spacegroups, etc. 

I think this should return a DataSet because otherwise you lose information about the Miller indices, spacegroup, cell, etc. Please let me know if you agree/disagree.

dennisbrookner:
I am going to make the changes above, and push them to a new branch `intensities-clean`. This is hopefully the easier path than trying to untangle the above PR.

dennisbrookner:
See #156

JBGreisman:
@dennisbrookner Can this be closed now that a new PR has been started?

dennisbrookner:
Yes, closing this

PR #156, opened 2022-05-27
Add function to convert structure factor amplitudes to intensities to `algorithms` submodule

dennisbrookner:
Now with the desired `DataSet`-based API.

I ended up deciding to name the new columns `I_back` and `SigI_back` by default, and to raise a `ValueError` if the method would overwrite a column. Happy to be overruled on one or both of those.

### DIFF
--- a/reciprocalspaceship/algorithms/intensity.py
+++ b/reciprocalspaceship/algorithms/intensity.py
@@ -1,7 +1,9 @@
+import warnings
+
 import numpy as np
 
 import reciprocalspaceship as rs
-import warnings
+
 
 def compute_intensity_from_structurefactor(
     ds,
@@ -14,7 +16,7 @@ def compute_intensity_from_structurefactor(
     """
     Back-calculate intensities and approximate intensity error estimates from
     structure factor amplitudes and error estimates
-    
+
     Intensity computed as I = SigF*SigF + F*F. Intensity error estimate
     approximated as SigI = abs(2*F*SigF)
 
@@ -41,7 +43,7 @@ def compute_intensity_from_structurefactor(
     DataSet
         DataSet with 2 additional columns corresponding to back-calculated
         intensities and intensity error estimates
-    
+
     """
 
     if not inplace:
@@ -56,32 +58,35 @@ def compute_intensity_from_structurefactor(
                 f"Input {ds.__class__.__name__} contains NaNs "
                 f"in columns '{F_key}' and/or '{SigF_key}'. "
                 f"Please fix these input values, or run with dropna=True"
-                )
-    
+            )
+
     if output_columns is None:
         output_columns = ["I_back", "SigI_back"]
-    I_key, SigI_key, = output_columns
+    (
+        I_key,
+        SigI_key,
+    ) = output_columns
 
     if I_key in ds.columns:
         raise ValueError(
             f"Input {ds.__class__.__name__} already contains column '{I_key}'."
             f"Try again and use the output_columns argument to pick a new"
             f"output column name."
-            )
+        )
     if SigI_key in ds.columns:
         raise ValueError(
             f"Input {ds.__class__.__name__} already contains column '{SigI_key}'."
             f"Try again and use the output_columns argument to pick a new"
             f"output column name."
-            )
-    
+        )
+
     # Initialize outputs
     ds[I_key] = 0.0
     ds[SigI_key] = 0.0
-    
+
     F_np = ds[F_key].to_numpy()
     SigF_np = ds[SigF_key].to_numpy()
-    
+
     I = SigF_np * SigF_np + F_np * F_np
     SigI = np.abs(2 * F_np * SigF_np)
 
--- a/reciprocalspaceship/algorithms/intensity.py
+++ b/reciprocalspaceship/algorithms/intensity.py
@@ -1,19 +1,19 @@
 import numpy as np
 
 import reciprocalspaceship as rs
+from reciprocalspaceship.decorators import inplace
 
-
+@inplace
 def compute_intensity_from_structurefactor(
     ds,
     F_key,
     SigF_key,
     output_columns=None,
-    dropna=True,
     inplace=False,
 ):
     """
-    Back-calculate intensities and approximate intensity error estimates from
-    structure factor amplitudes and error estimates
+    Compute intensities (I) and uncertainty estimates (SigI) from structure 
+    factor amplitudes (F) and their uncertainties (SigF) using error propagation
 
     Intensity computed as I = SigF*SigF + F*F. Intensity error estimate
     approximated as SigI = abs(2*F*SigF)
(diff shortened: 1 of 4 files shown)


reciprocalspaceship/algorithms/intensity.py
```diff
@@ -0,0 +1,96 @@
+import warnings
```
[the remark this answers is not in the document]
dennisbrookner: ah you got me, changed my mind and forgot about that

reciprocalspaceship/algorithms/intensity.py
```diff
@@ -0,0 +1,96 @@
...
+    output_columns=None,
+    dropna=True,
+    inplace=False,
+):
+    """
+    Back-calculate intensities and approximate intensity error estimates from
+    structure factor amplitudes and error estimates
+
+    Intensity computed as I = SigF*SigF + F*F. Intensity error estimate
+    approximated as SigI = abs(2*F*SigF)
```
JBGreisman, line 19: I don't love the expression "back-calculate" -- it implies a workflow that isn't quite necessary. I would just say something like "Compute intensities (I) and uncertainty estimates (SigI) from structure factor amplitudes (F) and their uncertainties (SigF) using error propagation."

reciprocalspaceship/algorithms/intensity.py
```diff
@@ -0,0 +1,94 @@
...
+        Column label for structure factor error estimates to use in
+        back-calculation
+    output_columns : list or tuple of column names
+        Column labels to be added to ds for recording back-calculated I and
+        SigI, respectively. output_columns must have len=2.
+    dropna : bool
+        Whether to drop reflections with NaNs in F_key or SigF_key
+        columns
+    inplace : bool
+        Whether to modify the DataSet in place or create a copy
```
JBGreisman, line 37: I was originally thinking that this function would always return a new DataSet that just had [F, SigF,  I, SigI] as the columns. This would avoid the need for `output_columns` and `inplace` arguments. 

Do you think there is a significant advantage to letting this function be performed `inplace` on a given DataSet? If so, this can be modified to use the `@inplace` decorator to avoid needing to implement it yourself -- see functions in `dataset.py` for example usage.

reciprocalspaceship/algorithms/intensity.py
```diff
@@ -0,0 +1,94 @@
...
+        Column label for structure factor amplitudes to use in back-calculation
+    SigF_key : str
+        Column label for structure factor error estimates to use in
+        back-calculation
+    output_columns : list or tuple of column names
+        Column labels to be added to ds for recording back-calculated I and
+        SigI, respectively. output_columns must have len=2.
+    dropna : bool
+        Whether to drop reflections with NaNs in F_key or SigF_key
+        columns
```
JBGreisman, line 35: Since this error propagation only depends on individual Miller indices, it seems to me that including NaNs shouldn't be problematic for the calculation. If there are NaNs in the input, there will be corresponding NaNs in the output. Am I missing something here? I'm not sure whether it's useful to sanitize the input here if it doesn't harm the calculation.

reciprocalspaceship/algorithms/intensity.py
```diff
@@ -0,0 +1,94 @@
...
+    Returns
+    -------
+    DataSet
+        DataSet with 2 additional columns corresponding to back-calculated
+        intensities and intensity error estimates
+
+    """
+
+    if not inplace:
+        ds = ds.copy()
```
JBGreisman, line 48: See comment above about `@inplace` decorator.

reciprocalspaceship/algorithms/intensity.py
dennisbrookner: I will run this through `pytest` to confirm but I believe you're right here

reciprocalspaceship/algorithms/intensity.py
```diff
@@ -0,0 +1,94 @@
...
+    ds[SigI_key] = 0.0
+
+    F_np = ds[F_key].to_numpy()
+    SigF_np = ds[SigF_key].to_numpy()
+
+    I = SigF_np * SigF_np + F_np * F_np
+    SigI = np.abs(2 * F_np * SigF_np)
+
+    ds[I_key] = rs.DataSeries(I, index=ds.index, dtype="Intensity")
+    ds[SigI_key] = rs.DataSeries(SigI, index=ds.index, dtype="Stddev")
```
JBGreisman, line 92: This can be greatly shortened in my mind:

```python
ds[I_key] = (ds[F_key]*ds[F_key] + ds[SigF_key]*ds[SigF_key]).astype("Intensity")
ds[SigI_key] = (2*ds[F_key]*ds[SigF_key]).abs().astype("Stddev")
```
Seems unnecessary in my mind to initialize the columns and convert to numpy arrays -- the operations are already vectorized within Pandas. Please let me know if you feel otherwise. I'm not sure if this was done to avoid any gotchas

reciprocalspaceship/algorithms/intensity.py
```diff
@@ -0,0 +1,94 @@
...
+    else:
+        if ds[[F_key, SigF_key]].isna().to_numpy().any():
+            raise ValueError(
+                f"Input {ds.__class__.__name__} contains NaNs "
+                f"in columns '{F_key}' and/or '{SigF_key}'. "
+                f"Please fix these input values, or run with dropna=True"
+            )
+
+    if output_columns is None:
+        output_columns = ["I_back", "SigI_back"]
```
JBGreisman, line 62: The `back` suffix is a bit confusing in my mind. Personally, I think it should be `I_calc` and `SigI_calc` instead of `back`. I think `back` implies that they were once intensities to begin with, but that doesn't have to be the case.

tests/algorithms/test_intensity.py
```diff
@@ -0,0 +1,52 @@
...
+    assert isinstance(result[o1].dtype, rs.IntensityDtype)
+    assert isinstance(result[o2].dtype, rs.StandardDeviationDtype)
+
+    # confirm arithmetic
+    F = result["F"].to_numpy()
+    SigF = result["SIGF"].to_numpy()
+    I = result[o1].to_numpy()
+    SigI = result[o2].to_numpy()
+    assert np.isclose(I, SigF * SigF + F * F).all()
+    assert np.isclose(SigI, np.abs(2 * F * SigF)).all()
```
JBGreisman, line 40: This is totally fine and we use it in the test suite in other places, but `np.allclose(...)` can be used here instead of `np.isclose(...).all()`

tests/algorithms/test_intensity.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
+import pytest
+
+import reciprocalspaceship as rs
+
+
+@pytest.mark.parametrize("inplace", [True, False])
+@pytest.mark.parametrize("output_columns", [None, ("I_backcalc", "SigI_backcalc")])
+def test_compute_intensity_from_structurefactor(ref_hewl, inplace, output_columns):
```
JBGreisman, line 9: Nice test -- if we decide to keep the `dropna` argument, please add a test (or parametrize this test) to hit that part of the function

JBGreisman (changes requested):
Thanks for updating this! It looks mostly good to me -- certainly the core functionality is here. I requested a few questions about the call signature, and some small code updates. Let me know if I can help out with any of these.

reciprocalspaceship/algorithms/intensity.py
JBGreisman: Ah,  you're right --  I had thought`rs.algorithms.scale_merged_intensities()` always returned a new DataSet. In that case, I think it's fine to do it this way for consistency. 

I still think the `@inplace` decorator should be used to provide that functionality -- the FW code predates that decorator, and it should be updated in the near future for consistency.

reciprocalspaceship/algorithms/intensity.py
```diff
@@ -0,0 +1,81 @@
...
+    # Sanitize input or check for invalid reflections
+    # if dropna:
+    #     ds.dropna(subset=[F_key, SigF_key], inplace=True)
+    # else:
+    #     if ds[[F_key, SigF_key]].isna().to_numpy().any():
+    #         raise ValueError(
+    #             f"Input {ds.__class__.__name__} contains NaNs "
+    #             f"in columns '{F_key}' and/or '{SigF_key}'. "
+    #             f"Please fix these input values, or run with dropna=True"
+    #         )
```
JBGreisman, line 53: Was this commented codeblock intended to make it into the last commit?

dennisbrookner: Ah rats, no that can go away for sure

JBGreisman:
Thanks for sticking with this -- it's looking good to me. Just two small comments:

1. It looks like a commented chunk of code was included in the last commit -- please remove that for the final commit. 
2. For thoroughness, it seems worthwhile to confirm that `F_key` and `SigF_key` select columns with the correct `dtype`, and to raise a descriptive `ValueError` if they don't.

dennisbrookner:
Checking for input types makes a lot of sense to me, but I'm not actually sure I know how to do that. Could you point me towards an example that exists elsewhere in `rs`?

JBGreisman:
Your test includes an example close to what I was thinking:
```python
# confirm types of new columns
assert isinstance(result[o1].dtype, rs.IntensityDtype)
assert isinstance(result[o2].dtype, rs.StandardDeviationDtype)
```
You can just change this to be something like:
```python
if not isinstance(ds.dtypes[F_key], rs.StructureFactorAmplitudeDtype):
    raise ValueError(...)
if not isinstance(ds.dtypes[SigF_key], rs.StandardDeviationDtype):
    raise ValueError(...)
```

tests/algorithms/test_intensity.py
```diff
@@ -0,0 +1,70 @@
...
+    # Function should require inputs of the correct types
+    with pytest.raises(ValueError):
+        rs.algorithms.compute_intensity_from_structurefactor(
+            ref_hewl, "F", "F", output_columns=("I", "SIGI")
+        )
+
+    with pytest.raises(ValueError):
+        rs.algorithms.compute_intensity_from_structurefactor(
+            ref_hewl, "SIGF", "SIGF", output_columns=("I", "SIGI")
+        )
```
JBGreisman, line 70: Don't include `output_columns=("I", "SIGI")` in these tests (can also move them to their own test if you'd like). As is, a ValueError is raised because of the `output_columns` values, which are checked before the type-checking occurs

JBGreisman:
All of the code looks good to me, but please double check your test of the `F_key`, `SigF_key` dtypes -- I don't think the test actually hits the relevant lines as written

tests/algorithms/test_intensity.py
dennisbrookner: Ah you're totally right, that's just careless copy/pasting on my part.

JBGreisman (approved):

Issue #103, opened 2021-10-13
MTZ data set hierarchy?


DHekstra:
I was curious whether it's possible to output MTZ files that support assigning 'crystals' and/or 'projects' to columns (in the sense of https://www.ccp4.ac.uk/html/mtzformat.html). I ran into this issue when trying to run a data set through SCALEIT and it objecting that data sets scaled belonged to the same crystal. I circumvented the issue by outputting separate mtz files and merging them in CAD. My understanding is that GEMMI can handle project names and crystal names (https://gemmi.readthedocs.io/en/latest/hkl.html#mtz-format).

JBGreisman:
This is something I've been considering implementing. It is supported in `gemmi`, so it wouldn't be challenging to at the very least allow custom naming of projects/crystals. 

Right now, I construct a `gemmi.Mtz` with a dataset named "reciprocalspaceship":
https://github.com/Hekstra-Lab/reciprocalspaceship/blob/f98b0dd0e9e665c76153e17837bd4ca123780f23/reciprocalspaceship/io/mtz.py#L103. 
I can add an attribute that allows one customize that name. 

However, I will note that I run `scaleit` pretty often and haven't had this issue, so I think we can also get around this issue with a change to your workflow.

PR #157, opened 2022-05-31
Add interface for setting project, crystal, and dataset names

JBGreisman:
This PR addresses #103 by adding an interface for setting the project, crystal, and dataset names in `DataSet.to_gemmi()` and `DataSet.write_mtz()`. These names are then set in the MTZ file or `gemmi.Mtz` object.

This can be useful for record keeping, allowing custom titles in MTZ files that can be read by other programs. 

These names default to "reciprocalspaceship" -- if these new arguments are left unset, a user will have the same behavior as before.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -319,9 +319,9 @@ class DataSet(pd.DataFrame):
     def to_gemmi(
         self,
         skip_problem_mtztypes=False,
-        project_name=None,
-        crystal_name=None,
-        dataset_name=None,
+        project_name="reciprocalspaceship",
+        crystal_name="reciprocalspaceship",
+        dataset_name="reciprocalspaceship",
     ):
         """
         Creates gemmi.Mtz object from DataSet object.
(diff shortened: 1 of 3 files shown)


kmdalton (changes requested):
I think this code would be cleaner and your tests more straightforward if the signature of `to_gemmi` were changed to

```python
def to_gemmi(
        self, 
        skip_problem_mtztypes=False, 
        project_name="reciprocalspaceship",
        crystal_name="reciprocalspaceship",
        dataset_name="reciprocalspaceship",
    ):
```
I _think_ strings are immutable, and this is okay to do. Let me know if I am being obtuse, and I will have another look.

JBGreisman:
I think that's a reasonable request. It certainly makes it more transparent what the default value will ultimately be.

kmdalton (changes requested):
I think that passing `DataSet.to_gemmi` anything other than a string for `{project,crystal,dataset}_name` should raise a `ValueError`. Passing `None` should probably not be silently corrected back to the default values in the signature. I think the tests should look for a `ValueError` when `None` is passed. I might be being overly picky.

JBGreisman:
I'm fine with that -- I agree that silently coercing a value of `None` to `"reciprocalspaceship"` is a bit surprising from an interface perspective. The docstrings certainly specify that the values take `str`, so I think it's reasonable to raise a `ValueError` for anything else.

kmdalton (changes requested):
I think that the methods `DataSet.to_gemmi` and `DataSet.write_mtz` should be the only places with default arguments for the various name strings. The functions in `io` are not really meant to be called directly by users. I think it is good for them to have the names as required arguments. This way, if we ever decide to change the default names in the future, they only need to be updated at the top level.

kmdalton (approved):
i don't see any glaring issues here, and i like the solution.

PR #160, opened 2022-06-06
Use gemmi method to implement `rs.utils.compute_structurefactor_multiplicity()`

JBGreisman:
This PR rewrites `rs.utils.compute_structurefactor_multiplicity()` to be implemented with methods provided by the `gemmi.GroupOps` class. Originally, we had implemented this on our own to vectorize the calculation; however, a vectorized version of this calculation is now provided in `gemmi`.


kmdalton (approved):
woo fewer lines to maintain.

PR #164, opened 2022-07-04
Add features to the add_rfree function

minhuanli:
1. Support ccp4 convention now: 0 is the test set and key is FreeR_flag
2. Update related test codes
3. Fix the previous bugs about updating more than the free flag column. Add some test codes to prevent such issues in the future.

### DIFF
--- a/reciprocalspaceship/utils/rfree.py
+++ b/reciprocalspaceship/utils/rfree.py
@@ -3,7 +3,7 @@ import numpy as np
 from reciprocalspaceship.dtypes import MTZIntDtype
 
 
-def add_rfree(dataset, fraction=0.05, bins=20, inplace=False):
+def add_rfree(dataset, fraction=0.05, bins=20, ccp4_convention=False, inplace=False):
     """
     Add an r-free flag to the dataset object for refinement.
     R-free flags are used to identify reflections which are not used in automated refinement routines.
@@ -17,6 +17,10 @@ def add_rfree(dataset, fraction=0.05, bins=20, inplace=False):
         Fraction of reflections to be added to the r-free. (the default is 0.05)
     bins : int, optional
         Number of resolution bins to divide the free reflections over. (the default is 20)
+    ccp4_conventiion: bool, optional
+        Default False to use Phenix(CNS/XPLOR) convention: !=0 is test set and key is "R-free-flags".
+        Set to True to use ccp4 convention: ==0 is test set and the key is "FreeR_flag".
+        See https://www.ccp4.ac.uk/html/freerflag.html#description for convention details.
     inplace : bool, optional
 
     Returns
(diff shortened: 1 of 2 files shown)


tests/utils/test_rfree.py
```diff
@@ -18,6 +18,24 @@ def test_add_rfree(self):
...
+
+        # Should keep all other unchanged
+        self.assertTrue(
+            np.all(data == rfree_ccp4.loc[:, rfree_ccp4.columns != "FreeR_flag"])
+        )
+        # Should have fewer ==0 as test set
+        self.assertTrue(
+            np.sum(rfree_ccp4.loc[:, "FreeR_flag"] == 0) / len(rfree_ccp4) < 0.15
+        )
+
```
JBGreisman, line 38: This is all fine, but it would be useful to move this test over to `pytest`, rather than keeping it in `unittest` (this is purely this way for historical reasons). Specifically, this would allow us to `parametrize` over `ccp4_convention` and `fraction`.

JBGreisman: let me know if you're up for including that refactor in this PR -- otherwise I can tackle it in another one.

reciprocalspaceship/utils/rfree.py
```diff
@@ -15,8 +15,12 @@ def add_rfree(dataset, fraction=0.05, bins=20, inplace=False):
...
     fraction : float, optional
         Fraction of reflections to be added to the r-free. (the default is 0.05)
-    bins : int, optional
-        Number of resolution bins to divide the free reflections over. (the default is 20)
+    ccp4_conventiion: bool, optional
+        Default False to use Phenix(CNS/XPLOR) convention:
+        1 is test set, 0 is working set, and key is "R-free-flags".
+        Set to True to use ccp4 convention:
+        0 is test set, 1 is working set, and the key is "FreeR_flag".
+        See https://www.ccp4.ac.uk/html/freerflag.html#description for convention details.
```
JBGreisman, line 23: I find this description a bit confusing -- do you mind first defining it: "Whether to use the CCP4 convention for R-free flags (Default: False). If True, use CCP4 convention... If False, use Phenix convention... "

tests/utils/test_rfree.py
minhuanli: Do you mean rewriting the rest function with `pytest.mark.parameterize` decorators? Sure, I can tackle it in this PR.

reciprocalspaceship/utils/rfree.py
minhuanli: I will make the change.

tests/utils/test_rfree.py
JBGreisman: yep -- using the `parametrize` decorators, and rewriting it to a function-based, pytest style, rather than the object-based `unittest` one. This move was made because we find the `pytest` style to be more concise and expressive, with less boiler plate than the `unittest` style.

minhuanli:
Made the change according to Jack's comments

JBGreisman (approved):

Issue #169, opened 2022-07-21
provide seed for reproducible creation of Rfree flags


dennisbrookner:
I just had the unfortunate experience of re-running a cell in a jupyter notebook and regenerating all of my R-free flags 😭. While that is of course totally my fault, I'm also sure I won't be the last to do this.

To me, the solution is to allow a user to seed the random number generator, and then get reproducible R-free flags.

e.g.
```python
def add_rfree(dataset, fraction=0.05, ccp4_convention=False, inplace=False):
    ...
    test_set = np.random.random(len(dataset)) <= fraction
```
would become something like

```python
def add_rfree(dataset, fraction=0.05, ccp4_convention=False, inplace=False, seed=None):
    ...
    rng = np.random.default_rng(seed) # this accepts None! So convenient!
    test_set = rng.random(len(dataset)) <= fraction
```

This has the side benefit of switching us over to the new version of `numpy` random number generators, which I'm just now learning about, but are a definite improvement.

JBGreisman:
I agree this sounds like a good idea, and is consistent with how users would expect this sort of consistency to be achieved. 

I think this would be a good "first issue" for someone to work on. In the meantime, I will note that setting the random seed for `numpy` prior to the call to `rs.utils.add_rfree()` should get you a reproducible result:

```python
In [1]: import numpy as np

In [2]: import reciprocalspaceship as rs

In [3]: ds = rs.read_mtz("tests/data/fmodel/9LYZ.mtz")

In [4]: def test_noseed(ds):
    ...:     flags1 = rs.utils.add_rfree(ds, fraction=0.5)["R-free-flags"]
    ...:     flags2 = rs.utils.add_rfree(ds, fraction=0.5)["R-free-flags"]
    ...:     return np.array_equal(flags1, flags2)
    ...: 

In [5]: def test_seed(ds):
    ...:     np.random.seed(100)
    ...:     flags1 = rs.utils.add_rfree(ds, fraction=0.5)["R-free-flags"]
    ...:     np.random.seed(100)
    ...:     flags2 = rs.utils.add_rfree(ds, fraction=0.5)["R-free-flags"]
    ...:     return np.array_equal(flags1, flags2)
    ...: 

In [6]: test_noseed(ds)
Out[6]: False

In [7]: test_seed(ds)
Out[7]: True
```

dennisbrookner:
I just want to note in case anyone finds this thread in the future that 
```python
np.random.seed(seed)
[...]

Issue #159, opened 2022-06-04
`rs.utils.copy_rfree()` should take `rfree_key` argument


dennisbrookner:
I just ran into a situation where I had to manually rename columns before using `rs.utils.copy_rfree()`, because the function does not accept a custom column name. I think this call should become:
```python
copy_rfree(dataset, dataset_with_rfree, rfree_key="R-free-flags", inplace=False)
```
`"R-free-flags"`, the Phenix naming convention, is a sensible default value for `rfree_key`.

I feel similarly, though less strongly, that `rs.utils.add_rfree()` should also accept the `rfree_key` argument. I think this is relevant to the changes proposed in #154; I'm raising this as a separate issue because it is a quicker fix that I can tackle first.

JBGreisman:
Yep -- this proposed change seems reasonable to me.

PR #170, opened 2022-07-27
Updated functionality for `add_rfree` and `copy_rfree`

dennisbrookner:
Making changes to `rs.utils.add_rfree` as per #169 and changes to `rs.utils.copy_rfree` as per #159 

## `add_rfree` takes `seed`

The `seed` is passed through to `np.random.default_rng(seed)`, such that running `add_rfree` again with the same `seed` results in the same flags. `seed` defaults to `None`, and conveniently `np.random.default_rng(None)` is a valid call.

A test was added to `test_add_rfree()` to confirm that flags are the same when a seed is used, and are different when a seed is not used.

## `copy_rfree` handles column names flexibly

The recent addition of the `ccp4_convention` parameter to `add_rfree` means that there are two possible names for an r-free column, `"R-free-flags"` (phenix convention) or `"FreeR_flag"` (ccp4 convention). `copy_rfree` will now by default search the provided `dataset_with_rfree` for a column named `"R-free-flags"`, then `"FreeR_flag"`, and throw a `ValueError` if neither is found.

The user can override this behavior via the `custom_rfree_key` argument if the r-free flag column uses any other naming convention.

A function `test_copy_rfree()` was added to test successfully copying in all combinations of column naming. A function `test_copy_rfree_errors()` was added to verify that `ValueError`s are thrown when appropriate.

### Old `unittest` code
There is an existing test for `copy_rfree` written with `unittest`. I believe that all key functionalities are tested in my new `pytest` code, making the `unittest` code redundant, but I left it in for now and labelled it in its docstring that it is a "legacy" version. Let me know if there's anything in the `unittest` code that I should additionally migrate, and/or if the `unittest` code should me removed altogether.

### DIFF
--- a/reciprocalspaceship/utils/rfree.py
+++ b/reciprocalspaceship/utils/rfree.py
@@ -52,7 +52,7 @@ def add_rfree(dataset, fraction=0.05, ccp4_convention=False, inplace=False, seed
     return dataset
 
 
-def copy_rfree(dataset, dataset_with_rfree, inplace=False, custom_rfree_key=None):
+def copy_rfree(dataset, dataset_with_rfree, inplace=False, rfree_key=None):
     """
     Copy the rfree flag from one dataset object to another.
 
@@ -64,7 +64,7 @@ def copy_rfree(dataset, dataset_with_rfree, inplace=False, custom_rfree_key=None
         A dataset with desired r-free flags.
     inplace : bool, optional
         Whether to operate in place or return a copy
-    custom_rfree_key : str, optional
+    rfree_key : str, optional
         Name of the column containing rfree flags in dataset_with_rfree.
         If None, dataset_with_rfree will be checked for column "R-free-flags"
         (phenix convention) then column "FreeR_flag" (ccp4 convention)
@@ -76,12 +76,11 @@ def copy_rfree(dataset, dataset_with_rfree, inplace=False, custom_rfree_key=None
     if not inplace:
         dataset = dataset.copy()
 
-    if custom_rfree_key is not None:
-        if custom_rfree_key not in dataset_with_rfree.columns:
+    if rfree_key is not None:
+        if rfree_key not in dataset_with_rfree.columns:
             raise ValueError(
-                f"""Supplied dataset_with_rfree contains no column {custom_rfree_key}"""
+                f"""Supplied dataset_with_rfree contains no column {rfree_key}"""
             )
-        rfree_key = custom_rfree_key
     elif "R-free-flags" in dataset_with_rfree.columns:
         rfree_key = "R-free-flags"
     elif "FreeR_flag" in dataset_with_rfree.columns:
--- a/reciprocalspaceship/utils/rfree.py
+++ b/reciprocalspaceship/utils/rfree.py
@@ -3,7 +3,7 @@ import numpy as np
 from reciprocalspaceship.dtypes import MTZIntDtype
 
 
-def add_rfree(dataset, fraction=0.05, ccp4_convention=False, inplace=False):
+def add_rfree(dataset, fraction=0.05, ccp4_convention=False, inplace=False, seed=None):
     """
     Add an r-free flag to the dataset object for refinement.
     R-free flags are used to identify reflections which are not used in automated refinement routines.
@@ -23,6 +23,10 @@ def add_rfree(dataset, fraction=0.05, ccp4_convention=False, inplace=False):
         1 is test set, 0 is working set, and key is "R-free-flags".
         See https://www.ccp4.ac.uk/html/freerflag.html#description for convention details.
     inplace : bool, optional
+    seed : int, optional
+        Seed to be passed to numpy.random.default_rng random number generator
+        for reproducible r-free flags. If None (default), r-free flags will
+        different each time.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


tests/utils/test_rfree.py
```diff
@@ -10,13 +10,18 @@
 @pytest.mark.parametrize("fraction", [0.05, 0.10, 0.15])
 @pytest.mark.parametrize("ccp4_convention", [False, True])
 @pytest.mark.parametrize("inplace", [False, True])
-def test_add_rfree(data_fmodel, fraction, ccp4_convention, inplace):
+@pytest.mark.parametrize("seed", [None, 2022])
+def test_add_rfree(data_fmodel, fraction, ccp4_convention, inplace, seed):
     """
-    Test rs.utils.add_rfee
+    Test rs.utils.add_rfree
```
JBGreisman, line 16: nice catch.

reciprocalspaceship/utils/rfree.py
```diff
@@ -46,7 +52,7 @@ def add_rfree(dataset, fraction=0.05, ccp4_convention=False, inplace=False):
     return dataset
 
 
-def copy_rfree(dataset, dataset_with_rfree, inplace=False):
+def copy_rfree(dataset, dataset_with_rfree, inplace=False, custom_rfree_key=None):
```
JBGreisman, line 55: I think this the parameter should be renamed to `rfree_key` -- I think the "custom" is unnecessary here.

JBGreisman (changes requested):
Thanks for taking care of these -- your changes look good to me. My only updates:
1. Please change `custom_rfree_key` --> `rfree_key`. I would rather keep the function arguments as short as possible, and I don't think the "custom" adds clarification. 
2. Please delete the now obsolete `unittest` code -- your `pytest` versions look good to me.

reciprocalspaceship/utils/rfree.py
```diff
@@ -66,8 +76,22 @@ def copy_rfree(dataset, dataset_with_rfree, inplace=False):
...
+            raise ValueError(
+                f"""Supplied dataset_with_rfree contains no column {rfree_key}"""
+            )
+    elif "R-free-flags" in dataset_with_rfree.columns:
+        rfree_key = "R-free-flags"
+    elif "FreeR_flag" in dataset_with_rfree.columns:
+        rfree_key = "FreeR_flag"
+    else:
+        raise ValueError(
+            """Failed to automatically find r-free flags in dataset_with_rfree. Please supply a custom_rfree_key"""
```
JBGreisman, line 90: sorry small thing -- do you mind changing `custom_rfree_key` here too in the ValueError?

dennisbrookner: Ah missed that, my bad

tests/utils/test_rfree.py
```diff
@@ -38,34 +43,72 @@ def test_add_rfree(data_fmodel, fraction, ccp4_convention, inplace):
...
+def test_copy_rfree_errors(data_fmodel):
+    """
+    Test expected ValueErrors for rs.utils.copy_rfree
+    """
+    with pytest.raises(ValueError):
+        rs.utils.copy_rfree(data_fmodel, data_fmodel)
 
-        return
+    with pytest.raises(ValueError):
+        rs.utils.copy_rfree(data_fmodel, data_fmodel, rfree_key="missing key")
```
JBGreisman, line 114: I didn't catch this before, but this can be simplified to:
```python
@pytest.mark.parametrize("rfree_key", [None, "missing key"])
def test_copy_rfree_errors(data_fmodel, rfree_key):
    """
    Test expected ValueErrors for rs.utils.copy_rfree()
    """
    with pytest.raises(ValueError):
        rs.utils.copy_rfree(data_fmodel, data_fmodel, rfree_key=rfree_key)
 ```
 I think the above is true... let me know if I'm mistaken

Comment (dennisbrookner) [reply tests/utils/test_rfree.py:114]: Oo good call, I think you're right. Didn't occur to me because it's a different `ValueError`, but yeah this should work.

Comment (dennisbrookner) [reply tests/utils/test_rfree.py:114]: I'll just clarify in the docstring what the two ValueErrors are, because I think the condensed version is a little harder to parse

Comment (dennisbrookner) [reply tests/utils/test_rfree.py]: In theory a slightly more rigorous test would be to confirm that the "missing key" case throws an error even when "R-free-flag" or "FreeR_flags" is present, but I think that's probably overkill.

Comment (JBGreisman) [reply tests/utils/test_rfree.py]: yep, I think your comment there is good. 

I think the test you just mentioned is a pretty good one because it tests the intended implementation (that `rfree_key` supercedes any other default column names). If we change the order of the `if` clauses in the future, that could change, so I think those sorts of tests are useful for making sure as many corner cases are covered.

File: tests/utils/test_rfree.py [hunk 5]
```diff
@@ -109,7 +109,7 @@ def test_copy_rfree_errors(data_fmodel):
         rs.utils.copy_rfree(data_fmodel, data_fmodel)
 
     # Raise ValueError because "missing key" is missing,
-    # even though "R-free-flags" is present
+    # even though "R-free-flags" exists
```
dennisbrookner, line 112: Had to make a change to get out of a loop of git madness, figured this was benign enough

JBGreisman (approved):
Looks good -- thanks for making the fixes!

PR #171, opened 2022-08-01
Handle symmetry ops that produce fractional HKLs in `apply_to_hkl()`

### LINKED ISSUE #134: Adding a warning to apply_symop()
When converting from a centered spacegroup like C2 to a non-centered one like P1, it is natural to apply a symop like
`gemmi.Op("1/2*x-1/2*y,1/2*x+1/2*y,z")`. When not careful with the input Miller indices, this can result in fractional Miller indices. It may be good to add a check that new indices are integers.

JBGreisman:
This PR fixes #134 by adding some additional logic to `rs.utils.apply_to_hkl()`. Before, the use of `np.floor_divide()` would return the closest integer HKL, regardless of whether the true symmetry operation yielded a fractional result. Now, symmetry operations that produce non-integer output on the given Miller indices raise a `RuntimeError`

### DIFF
--- a/reciprocalspaceship/utils/symop.py
+++ b/reciprocalspaceship/utils/symop.py
@@ -14,10 +14,29 @@ def apply_to_hkl(H, op):
 
     Returns
     -------
-    result : array
+    result : np.ndarray(int32)
         n x 3 array of Miller indices after operator application
+
+    Raises
+    ------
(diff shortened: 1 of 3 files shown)


kmdalton (approved):
I can't find any bugs.

Issue #31, opened 2020-12-15
Align Phases of Isomorphous Structures


kmdalton:
Often times, I notice that SAD phasing solutions from isomorphous structures don't overlay in real space. This seems to have something to do with the choice of "phase origin". It'd be super useful if we had a function to make sure that two sets of structure factors have compatible origins. We should be able to implement something akin to @apeck12's [method](https://github.com/apeck12/cryoetX/blob/2c9f7bd69b91cb9a71d2e90a05350c0af87c4770/ProcessCrystals.py#L459). After talking to @JBGreisman, we decided doing this in cases with good phases might boil down to solving a simple linear program. At any rate, this should absolutely be something we put in `rs.algorithms`.

apeck12:
@kmdalton Is the issue that both structures look correct but are rotated and translated with respect to each other? If that’s the case, then a full phase origin search isn’t needed (except in P1). There’s a limited number of equivalent phase origins for each space group — P212121, for example, has only 4 — so one could get away with just mapping the phases of the second structure onto each of those equivalent origins and checking for consistency with the first.

kmdalton:
Yes, I think you've characterized the issue correctly, @apeck12. Do you know how we can generate the phase origins for a particular group programmatically? In cases like P21 or P61 without orthogonal symmetry elements, are there still finitely many phase origins, or are they restricted to a line?

kmdalton:
I can confirm that 

1.  Searching a finite set of origins works to align phases in tetragonal lysozyme. I have started work on a new [branch](https://github.com/Hekstra-Lab/reciprocalspaceship/blob/phalign/reciprocalspaceship/algorithms/phase.py) that implements the algorithm I tested. 
2.  Arbitrarily translating along the z-axis in spacegroup [P63](http://img.chem.ucl.ac.uk/sgp/large/173az1.htm) leads to valid phase combinations and nice electron density. 
3.  Arbitrarily translating the phases along other axes does not result in interpretable density. 

Based on this, I think we need a hybrid approach to cover all space groups. Space groups with orthogonal symmetry axes should be handled by the approach in 1. I'm afraid we will need to learn the translation vector by a grid search and/or optimization in cases like P63. 

What edge cases am I missing? 

How can we write effective tests for these new algorithms?

apeck12:
A few thoughts:

[...]

PR #174, opened 2022-08-23
Model-free alignment of isomorphous maps (alternative origins problem)

JBGreisman:
This PR is still in progress, but I wanted to upload my initial classes to support this. The problem is broken down into two cases:
1. Non-Polar spacegroups: easier case because they have a finite number of cases to evaluate
2. Polar spacegroups: more challenging because of the continuous alternative origins. 

These classes use the sum of the squared phase residual between two sets of phases as a loss function to minimize, which has been pretty effective in my hands. 

The `NonPolarTranslator` does a brute-force search of the 64 possible cases to evaluate the optimal translation vector. The `PolarTranslator` does a global optimization to determine the optimal translation vector. This is a bit non-trivial -- its a bumpy loss landscape -- but I've found the scipy dual_annealing optimizer to work very well. Another trick has been to start at low-resolution and gradually increase to full resolution. After discussion with @kmdalton, we can also use a FFT-based method to come up with a sensible initial value for optimization. 

My todo items:
- [ ] Add user-facing method that takes two datasets, and uses the proper class based on the spacegroup to determine the translation vector that gives a common phase origin.
- [ ] Add a suite of tests for common spacegroups that cover the polar/non-polar cases
- [ ] (optional improvement) Add initialization code based on FFT-method to speed things up

### DIFF
--- a/reciprocalspaceship/algorithms/translator.py
+++ b/reciprocalspaceship/algorithms/translator.py
@@ -1,0 +1,149 @@
+import numpy as np
+
+
+class BaseTranslator:
+    def __init__(self, H, phi_ref, phi, dhkl):
+        """
+        Translator class for determining the optimal translation
+        vector to align two sets of phases.
+
+        Parameters
+        ----------
+        H : np.ndarray
+            Miller indices as an N x 3 array
+        phi_ref : np.ndarray
+            Reference phases in degrees (length N)
+        phi : np.ndarray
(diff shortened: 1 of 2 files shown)


JBGreisman:
As per the comment in #31, this PR seems better suited for the `rs-booster` project because this is an application that uses `rs`, rather than a new core feature of the library. I am closing this PR here, and will add this via a PR in `rs-booster`.

PR #176, opened 2022-08-24
Add `is_polar()` utility for classifying polar space groups

JBGreisman:
Aligning maps of isomorphous structures can benefit from classifying polar space groups. This is useful for distinguishing which cases have a discrete set of alternative origins vs. continuous alternatives (see #174).

This PR implements a `reciprocalspaceship.utils.is_polar()` utility function that classifies a space group as polar. This function is tested against the 530 spacegroup settings in `sgtbx`.

### DIFF
--- a/tests/utils/test_polar.py
+++ b/tests/utils/test_polar.py
@@ -1,0 +1,44 @@
+from os.path import abspath, dirname, join
+
+import gemmi
+import pandas as pd
+import pytest
+import reciprocalspaceship as rs
+
+
+def sgtbx_polar_classification():
+    """
+    Helper function for generating Hermann-Mauguin (xhm) symbols with
+    corresponding polar classifications
+    """
+    data = ["..", "data", "sgtbx", "sgtbx_polar.csv"]
+    inFN = abspath(join(dirname(__file__), *data))
+    ref = pd.read_csv(inFN)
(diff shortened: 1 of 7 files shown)


tests/utils/test_polar.py
```diff
@@ -0,0 +1,45 @@
...
+    corresponding polar classifications
+    """
+    data = ["..", "data", "sgtbx", "sgtbx_polar.csv"]
+    inFN = abspath(join(dirname(__file__), *data))
+    ref = pd.read_csv(inFN)
+    return ref.iterrows()
+
+
+@pytest.fixture(params=sgtbx_polar_classification())
+def polar_by_xhm(request):
```
kmdalton, line 22: admittedly, i do not understand fixtures all that well, but i found this file surprisingly difficult to read. i think this might be helped by adding 
```
Parameters
---------------
request : ??????
```
to the docstring.

kmdalton (changes requested):
I think this looks very good. I don't see anything objectionable except that I found the test really hard to understand. I think that is on me not understanding pytest that well, but maybe you can spruce up the docstring to help others understand it in the future.

kmdalton (approved):
I still find this hard to understand, but I am not going to hold you responsible for pytest's API choices.

PR #177, opened 2022-08-25
Add `rs.utils.polar_axes()` for classifying which cell axes permit continuous origin shifts

JBGreisman:
Based on the routine implemented in `is_polar()`, it's also possible to return which of the cell axes in a given spacegroup are polar (permit continuous origin shifts). This is implemented here as `rs.utils.polar_axes()` with the following call signature:

`polar_axes(spacegroup) --> List(3) of bools`

For example:
```python 
print(rs.utils.polar_axes("P 1"))         # [True, True, True]
print(rs.utils.polar_axes("P 1 21 1"))    # [False, True, False]
print(rs.utils.polar_axes("P 21 21 21"))  # [False, False, False]
```

One implementation decision was to only support hexagonal Bravais lattices for the relevant "R 3" trigonal space groups. This decision is made so that it is clear that the function returns polar axes corresponding to the a, b, and c-axes, and not the a', b', c' corresponding to the rhombohedral setting. The supported hexagonal case is the reference setting for these spacegroups, so I think this is a reasonable design choice. To avoid confusion, a ValueError is raised when such space groups are given with a rhombohedral setting. 

The code is tested against all applicable spacegroup settings in `sgtbx` using the `is_allowed_origin_shift()` method.

### DIFF
--- a/reciprocalspaceship/utils/symmetry.py
+++ b/reciprocalspaceship/utils/symmetry.py
@@ -98,8 +98,8 @@ def polar_axes(spacegroup):
 
     Returns
     -------
-    list
-        List of whether a-, b-, or c-axis is polar
+    List[bool]
+        Whether a-, b-, or c-axis is polar
 
     Raises
     ------
--- a/reciprocalspaceship/utils/symmetry.py
+++ b/reciprocalspaceship/utils/symmetry.py
@@ -78,3 +78,41 @@ def is_polar(spacegroup):
     sym_ops = spacegroup.operations().sym_ops
     a = np.array([op.rot for op in sym_ops])
     return ~(a < 0).any(axis=1).any(axis=0).all()
+
+
+@spacegroupify
+def polar_axes(spacegroup):
+    """
+    Classify which cell axes are polar in given spacegroup
+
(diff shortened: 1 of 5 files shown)


reciprocalspaceship/utils/symmetry.py
```diff
@@ -78,3 +78,41 @@ def is_polar(spacegroup):
...
+
+    Parameters
+    ----------
+    spacegroup : str, int, gemmi.SpaceGroup
+        Spacegroup to classify axes as polar
+
+    Returns
+    -------
+    list
+        List of whether a-, b-, or c-axis is polar
```
kmdalton, line 102: make clear this is a list of booleans.

kmdalton (changes requested):
Just a small nitpick. About a docstring.

reciprocalspaceship/utils/symmetry.py
JBGreisman: makes sense -- updated in latest commit

kmdalton (approved):
You have resolved my issues. Looks good!

Issue #175, opened 2022-08-23
Add `sample_rate` or `resolution_cutoff` options for `to_reciprocalgrid`


kmdalton:
Right now you need to specify the number of grid points along h, k, and l when creating a reciprocal space grid of a column in a data set. This is good for flexibility, but it is not very convenient. It would be nice to have the alternative to specify a `sample_rate` as in `gemmi` or a `resolution_cutoff`.

dennisbrookner:
As it happens, I had just written this little helper function as a workaround:
```python
def spacing_to_gridsize(spacing, cell):
    """
    Compute the optimal gridsize based on unit cell size and desired spacing

    spacing : float
        Desired (approximate) grid spacing in Angstroms
    cell : gemmi.UnitCell
        Or anything similar with attributes a, b, and c

    NOTE: does not support different spacing in each direction, but in theory could
    """
    gridsize = []
    for dim in [cell.a, cell.b, cell.c]:

        gridsize.append(int(dim // spacing))

    return gridsize
```

JBGreisman:
I think the most reliable implementations would use `gemmi` for this:

resolution cutoff:
```
In [1]: ds.cell.get_hkl_limits(0.95) # Grid size should be this *2 to handle negative values
Out[1]: [28, 33, 36]
```
sampling rate:
```
In [2]: ds.to_gemmi().get_size_for_hkl(sample_rate=3)
Out[2]: [80, 96, 108]
```

kmdalton:
Just a note. `sample_rate` and `resolution_cutoff` / `dmin` both influence grid size but are not mutually exclusive concepts. I think we should support both simultaneously so that

```python
ds.to_reciprocalgrid("F", sample_rate=3., dmin=5.)
```

produces a 3x oversampled grid from the reflections out to 5A. 

Does that make sense? Am I talking crazy?

JBGreisman:
I think that's reasonable. That latter case can still be implemented easily in gemmi for first thresholding at the given `dmin` prior to the `to_gemmi()` call. 

Just to summarize the proposed API change, there will be two mutually exclusive sets of options:

1. `DataSet.to_reciprocalgrid(grid_size=(int, int, int))`
2. `DataSet.to_reciprocalgrid(sample_rate=float)` or `DataSet.to_reciprocalgrid(dmin=float)` or `DataSet.to_reciprocalgrid(sample_rate=float, dmin=float)`

kmdalton:
[...]

PR #179, opened 2022-08-30
Add `sample_rate` and `dmin` arguments to `DataSet.to_reciprocalgrid()`

JBGreisman:
This PR addresses #175 by updating the `DataSet.to_reciprocalgrid()` call signature to accept `sample_rate` and `dmin` arguments. The default value for `sample_rate=3.0` and the default for `dmin=None`, which corresponds to using the full resolution of the `DataSet` instance. The `to_reciprocalgrid()` method can still accept an explicit `gridsize`,  and if one is provided, it supersedes the `sample_rate` and `dmin` arguments. 

To facilitate this method, I also added `rs.utils.get_gridsize()`, which can be used for calculating a proper gridsize for a `DataSet` instance. 

Lastly, this PR cleans up and adds additional tests to `test_dataset_grid.py` to more thoroughly test this function.


kmdalton:
I think a `get_fourier_gridsize` type of function should exist in `utils` because it seems generally useful especially in the context of initializing deep learning models. I don't think _this_ `get_gridsize` function belongs in `utils` though. In my mind nothing in `utils` should require a `dataset` instance. Let's do this at the top level for now. If you want a separate method to handle the grid sizing, I'd suggest writing a "private" `DataSet._get_gridsize` helper. 

I can do a more thorough review tomorrow.

JBGreisman:
Thanks for the comment -- I agree that as written the function wasn't right for `rs.utils`, which we try to keep agnostic to any implementation/interface details in DataSet. 

I just pushed an update to this PR that changes the `rs.utils.get_gridsize()`call signature to be more general. I also added a specific helper function `DataSet.get_gridsize()`, which is then called in `DataSet.to_reciprocalgrid()`.

reciprocalspaceship/dataset.py
```diff
@@ -1287,31 +1287,83 @@ def canonicalize_phases(self, inplace=False):
...
-        Miller indices. A 3D numpy array will be initialized with the
-        specified gridsize. Any missing Miller indices are initialized to
-        zero.
+        Miller indices.
 
         Notes
         -----
         - The data being arranged on a reciprocal grid must be compatible
           with a numpy datatype.
+        - Any missing Miller indices are initialized to zero.
```
kmdalton, line 1329: This is out of scope for this PR, but we should consider initializing by interpolating adjacent grid points. Along these lines though, should there be a `fill_value=0.` option?

reciprocalspaceship/dataset.py
```diff
@@ -1287,31 +1287,83 @@ def canonicalize_phases(self, inplace=False):
...
+        - The grid size determined using `sample_rate` and `dmin` will depend
+          on the cell parameters of the dataset. If the grid size must be
+          consistent across different isomorphous cell parameters, `gridsize`
+          can be explicitly provided.
```
kmdalton, line 1333: explicitly provided. `gridsize` supersedes `dmin` and `sample_rate` for grid size determination.

reciprocalspaceship/dataset.py
```diff
@@ -1287,31 +1287,83 @@ def canonicalize_phases(self, inplace=False):
...
         key : str
             Column label for value to arrange on reciprocal grid
+        sample_rate : float
+            Sets the minimal grid spacing relative to dmin. For example,
+            `sample_rate=3` corresponds to a real-space sampling of dmin/3.
+            (default: 3.0)
+        dmin : float
+            Highest-resolution reflection to consider for grid size. If None,
+            `dmin` will be set to the highest resolution reflection in the
+            dataset (default: None)
```
kmdalton, line 1346: Need to say: "The reflections used to populate the grid will also be truncated to resolution `>=dmin`. "

reciprocalspaceship/dataset.py
```diff
@@ -1287,31 +1287,83 @@ def canonicalize_phases(self, inplace=False):
 
         return self
 
-    def to_reciprocalgrid(self, key, gridsize):
+    def get_gridsize(self, dmin=None, sample_rate=3.0):
```
kmdalton, line 1290: I propose to change the name of this method to `get_fourier_gridsize` or `get_reciprocal_gridsize` something similar. Otherwise, it is not apparent what the "grid" is. I think the is the same reason it's called `ds.to_reciprocalgrid` and not simply `ds.to_grid`.

kmdalton: This is a nitpick but i think `gridsize` should be `grid_size` in parameter and method names. I don't see why we would contract it.

reciprocalspaceship/utils/grid.py
```diff
@@ -0,0 +1,55 @@
+import gemmi
+import numpy as np
+
+from reciprocalspaceship.decorators import cellify, spacegroupify
+
+
+@cellify
+@spacegroupify
+def get_gridsize(cell, dmin, sample_rate=3.0, spacegroup=None):
```
kmdalton, line 9: I'd prefer `get_fourier_grid_size` or `get_reciprocal_grid_size`

reciprocalspaceship/utils/grid.py
```diff
@@ -0,0 +1,55 @@
...
+        (default: None)
+
+    Returns
+    -------
+    list(int, int, int)
+        Grid size with desired spacing (list of 3 integers)
+
+    Raises
+    ------
+    ValueError
```
kmdalton, line 40: Nice. I always forget to do this.

tests/test_dataset_grid.py
```diff
@@ -5,49 +5,71 @@
...
+        result = dataset.to_reciprocalgrid("FMODEL", gridsize=gridsize)
+        assert np.allclose(result, expected)
 
 
-@pytest.mark.parametrize("sample_rate", [2.5, 3, 5])
-@pytest.mark.parametrize("p1", [True, False])
-def test_to_reciprocalgrid_float(mtz_by_spacegroup, sample_rate, p1):
+@pytest.mark.parametrize("sample_rate", [2.0, 3.0])
+@pytest.mark.parametrize("column", ["FMODEL", "sf"])
+def test_to_reciprocalgrid_p1(mtz_by_spacegroup, sample_rate, column):
```
kmdalton, line 41: Not sure this test is necessary, but I guess it doesn't hurt to have it.

tests/test_dataset_grid.py
```diff
@@ -5,49 +5,71 @@
...
+    result1 = dataset.to_reciprocalgrid(column, gridsize=gridsize)
+    result2 = p1.to_reciprocalgrid(column, gridsize=gridsize)
+    assert np.allclose(result1, result2)
 
-    assert np.allclose(result, expected)
+
+@pytest.mark.parametrize("sample_rate", [1.0, 2.5, 3.0])
+@pytest.mark.parametrize("dmin", [10.0, 5.0, None])
+@pytest.mark.parametrize("gridsize", [None, [144, 144, 144]])
+def test_to_reciprocalgrid_sizes(mtz_by_spacegroup, sample_rate, dmin, gridsize):
```
kmdalton, line 61: Looks good.

tests/utils/test_grid.py
```diff
@@ -0,0 +1,42 @@
...
+    if sample_rate < 1.0:
+        with pytest.raises(ValueError):
+            rs.utils.get_gridsize(dataset.cell, dmin=dmin, sample_rate=sample_rate)
+
+    else:
+        result = rs.utils.get_gridsize(dataset.cell, dmin=dmin, sample_rate=sample_rate)
+
+        # shape of output
+        assert isinstance(result, list)
+        assert len(result) == 3
```
kmdalton, line 27: as long as we're doing this, you may as well make sure the list contains ints. something like: 
```python
assert all([isinstance(i, int) for i in result])
```

kmdalton (changes requested):
I have some minor suggestions and some semantic concerns. Mostly I think anywhere that says `gridsize` should be `grid_size`, and we should make it clear what grid we're talking about in the method names. Either `get_grid_size` --> `get_fourier_grid_size` or `get_reciprocal_grid_size`. 

Otherwise, this seems like a great PR. As far as I can tell, the tests are well engineered.

reciprocalspaceship/dataset.py
JBGreisman: updated docstring

reciprocalspaceship/dataset.py
JBGreisman: updated docstring

reciprocalspaceship/dataset.py
JBGreisman: updated call signature and method name

tests/test_dataset_grid.py
JBGreisman: This test is very similar to the ones done for `expand_to_p1()`. Though somewhat repetitive, it seemed like a good idea to ensure that there is no dependency on `sample_rate`

tests/utils/test_grid.py
JBGreisman: good idea -- just added this.

reciprocalspaceship/utils/grid.py
JBGreisman: done -- for consistency elsewhere, I've made it `get_reciprocal_grid_size`

reciprocalspaceship/dataset.py
JBGreisman: I agree this could be a future change. Given how much is going on in this PR, I'd rather defer any additional changes

JBGreisman:
I made the requested call signature and method name changes -- I agree that they all clarify things, and make things more consistent with our naming scheme. 

The big additional API change is that I changed `DataSet.to_reciprocalgrid()`--> `DataSet.to_reciprocal_grid()` to make things as consistent as possible. Since this change will break backwards compatibility, I have kept `DataSet.to_reciprocalgrid(key, gridsize=...)` and added a `DeprecationWarning`. I have also updated the documentation with a warning, and have changed the example notebooks to use the new function. In a future release we can fully remove support for the deprecated version.

reciprocalspaceship/dataset.py
kmdalton: Fine with me.

reciprocalspaceship/dataset.py
```diff
@@ -3,7 +3,7 @@
 import pandas as pd
 from pandas.api.types import is_complex_dtype
 
-from reciprocalspaceship import concat, dtypes
+from reciprocalspaceship import concat, dtypes, utils
```
kmdalton, line 6: Why don't you just import `get_reciprocal_grid_size` in the `utils` function imports below? Wouldn't that be more consistent?

kmdalton (changes requested):
I think we're nearing completion here. See my comment about imports. 

In example 5 from the docs there is still a reference to `to_reciprocalgrid` in one of the markdown cells. You also use `reciprocalgrid` as a variable name in that example. You might want to update that, but that is pretty much just aesthetic. 

I couldn't find any other issues with misusing `grid_size` or `reciprocalgrid`.

JBGreisman:
Made the two changes -- thanks for the review! Please feel free to merge if you think this is good to go

kmdalton:
Sorry -- I forgot to point out that the call to `get_reciprocal_grid_size` still had a `utils` prefix. I just pushed a commit to fix. The CI should pass now :crossed_fingers:

JBGreisman:
Thanks -- good catch.

PR #182, opened 2022-09-06
`DataSet.to_numpy()` should use numpy dtypes whenever possible

### LINKED ISSUE #33: Invoking to_numpy on 2D array returns object dtype
Sometimes when converting multiple columns to `numpy`, `reciprocalspaceship` converts the columns to `object dtype`.

A simple example to reproduce:
```python
import numpy as np


ds = rs.DataSet({
    "X" : np.random.random(100),
    "Y" : np.random.random(100),
}).infer_mtz_dtypes()

print(f"ds.dtypes:\n{ds.dtypes}\n")
print(f"ds['X'].to_numpy().dtype:\n  {ds['X'].to_numpy().dtype}\n")
print(f"ds['Y'].to_numpy().dtype:\n  {ds['Y'].to_numpy().dtype}\n")
print(f"ds[['X', 'Y']].to_numpy().dtype:\n  {ds[['X', 'Y']].to_numpy().dtype}")
```
outputs:
```
ds.dtypes:
X    MTZReal
Y    MTZReal
dtype: object

ds['X'].to_numpy().dtype:
  float32

ds['Y'].to_numpy().dtype:
  float32

ds[['X', 'Y']].to_numpy().dtype:
  object
```

JBGreisman:
Pandas DataFrames that contain ExtensionDtypes always default to output data with `object` dtype when `DataFrame.to_numpy()` is called. This is suboptimal for MTZ data, which by construction must be compatible with `float32`, and possibly `int32`. 

This PR wraps the pandas call with `DataSet.to_numpy()` to assess whether a more sensible default (either `float32` or `int32`) can be used based on the existing data. This should help to avoid cases where data is unnecessarily cast to an `object` array, which can lead to unexpected behavior downstream.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -1,6 +1,9 @@
+from inspect import signature
+
 import gemmi
 import numpy as np
 import pandas as pd
+from pandas._libs import lib
 from pandas.api.types import is_complex_dtype
 
 from reciprocalspaceship import concat, dtypes
@@ -11,7 +14,7 @@ from reciprocalspaceship.decorators import (
     range_indexed,
     spacegroupify,
 )
-from reciprocalspaceship.dtypes.base import MTZDtype
+from reciprocalspaceship.dtypes.base import MTZDtype, MTZInt32Dtype
 from reciprocalspaceship.utils import (
     apply_to_hkl,
     bin_by_percentile,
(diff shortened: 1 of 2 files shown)


reciprocalspaceship/dataset.py
```diff
@@ -292,6 +293,53 @@ def _handle_cached_dtypes(dataset, columns, drop):
             dataset = _handle_cached_dtypes(dataset, columns, drop)
             return dataset
 
+    def to_numpy(self, dtype=None, copy=False, na_value=lib.no_default):
+        """
+        Convert the DataSet to a NumPy array.
+
+        By default, the returned array will be `float32` if all columns of the DataSet are
```
kmdalton, line 300: I find this paragraph hard to read. I took a stab at improving it:

```
This method will attempt to infer a consensus numpy dtype from 
the dtypes of the DataSet columns. If the DataSet is composed of 
all `int32`-backed MTZ dtypes and does contain NaN values, the returned
 `dtype` will be `int32`. For all other combinations of `MTZDtype`, 
the returned dtype will be `float32`.  If the DataSet contains dtypes other 
than `MTZDtype`, the default Pandas behavior is used. 
```

tests/dtypes/test_dataset_to_numpy.py
```diff
@@ -0,0 +1,110 @@
...
+def test_to_numpy_ints(dtype, hasna):
+    """
+    Test DataSet.to_numpy() with int32-backed MTZDtypes
+    """
+    ds = rs.DataSet(
+        {
+            "X": np.random.randint(0, 100, size=100),
+            "Y": np.random.randint(0, 100, size=100),
+        }
+    ).infer_mtz_dtypes()
```
kmdalton, line 19: I for one find `infer_mtz_dtypes` to be a useful function, but I don't think you should rely on it for the purposes of this test. If the behavior changes in the future, it will make for some annoying debugging. Explicitly set these dtypes (likewise for the other tests).

tests/dtypes/test_dataset_to_numpy.py
```diff
@@ -0,0 +1,110 @@
...
+    if dtype is None:
+        # Fall back to float32
+        assert arr.dtype == np.float32
+    else:
+        assert arr.dtype == dtype
+
+
+@pytest.mark.parametrize("dtype", [None, np.float32, np.float64, object])
+@pytest.mark.parametrize("hasna", [True, False])
+def test_to_numpy_mtzdtypes(dtype, hasna):
```
kmdalton, line 69: This test name is a little confusing, the first 3 tests all use `mtzdtypes`. Maybe `test_to_numpy_mixed_mtzdtypes` or something. Likewise maybe the previous tests should be `test_to_numpy_mtzints` and `test_to_numpy_mtzfloats`.

kmdalton (changes requested):
I like it. I wish we also handled the case where `MTZDtype` is mixed with Pandas `float` dtypes. I'm not sure that can be implemented in a way that doesn't cause nasty side effects.

reciprocalspaceship/dataset.py
```diff
@@ -292,6 +293,55 @@ def _handle_cached_dtypes(dataset, columns, drop):
             dataset = _handle_cached_dtypes(dataset, columns, drop)
             return dataset
 
+    def to_numpy(self, dtype=None, copy=False, na_value=lib.no_default):
+        """
+        Convert the DataSet to a NumPy array.
```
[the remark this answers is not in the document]
JBGreisman: added link to main paragraph of docstring

reciprocalspaceship/dataset.py
JBGreisman: Updated to use this language. I also added a link to the Pandas documentation at the end.

tests/dtypes/test_dataset_to_numpy.py
JBGreisman: Fair point -- these tests should be minimally dependent on other features. I've updated the relevant clauses to explicitly set dtypes

tests/dtypes/test_dataset_to_numpy.py
JBGreisman: I updated the test function names to be a bit more explicit

kmdalton (approved):

Issue #173, opened 2022-08-05
Resolving indexing ambiguities when comparing datasets


JBGreisman:
One of the goals of `reciprocalspaceship` is to make it easier to compare datasets in reciprocal space. For some spacegroups, indexing ambiguities can confound such comparisons. We should add some functions to `rs` to make it easier to deal with these spacegroups. 

In particular, we can use `gemmi.find_twin_laws()` to identify possible reindexing operations, and we can use correlation coefficients to determine what the best operation is. The gemmi function seems to produce output consistent with the [CCP4 table (poor formatting)](https://www.ccp4.ac.uk/html/reindexing.html#lookup-tables), but I've only tested a handful of cases:

```python
# PYP
In [1]: gemmi.find_twin_laws(gemmi.UnitCell(66.9, 66.9, 40.8, 90, 90, 120), gemmi.SpaceGroup("P 63"), max_obliq=1.0, all_ops=False)
Out[1]: [<gemmi.Op("-x,-x+y,-z")>]

# PTP1B
In [2]: gemmi.find_twin_laws(gemmi.UnitCell(89.3, 89.3, 105.7, 90, 90, 120), gemmi.SpaceGroup("P 31 2 1"), max_obliq=1.0, all_ops=False)
Out[2]: [<gemmi.Op("-x,-y,z")>]

```

I think this should involve two additions to `rs`:
1. Helper function for `rs.DataSet` that returns possible reindexing operations (perhaps `rs.utils.possible_reindexing_ops(dataset)`). This function could then be used elsewhere, wherever such a functionality would be desired. 
2. `rs.algorithms.resolve_indexing_ambiguity(dataset1, dataset2, column_key="IMEAN")`, which applies the reindexing operation to dataset2 that gives the best correlation in IMEAN. 

Please let me know if anyone has any thoughts on this -- together with #31, this should handle some of the gotchas when making comparisons across isomorphous structures/datasets

kmdalton:
I think this is a great feature to add. I'm not crazy about the function names. I think we could workshop them a little. Here are alternative signatures off the top of my head. Not saying these are perfect either. 
 1) `rs.DataSet.reindexing_ops()` <-- I see this as more of a bound method thing. the `utils` method should just take a `cell` and `spacegroup` to be consistent with the rest of the `utils` api. However, the gemmi method already covers this. In my mind, there is no need to wrap it at the `utils` level. Thoughts?
[...]

PR #183, opened 2022-09-07
Add `DataSet.reindexing_ops` and `DataSet.find_twin_laws()`

JBGreisman:
As discussed in #173, it is useful to have ways to get merohedral twin laws, which are possible reindexing operations for resolving ambiguities. In other analysis tasks, it might be useful to get the full set of possible pseudo-merohedral twin laws for a `DataSet` with its spacegroup/cell.

This PR adds an attribute (`DataSet.reindexing_ops`) to access merohedral twin laws and a method (`DataSet.find_twin_laws()`) which can be used to access the full list of possible twin laws. These are implemented using calls to `gemmi.find_twin_laws()` and are tested against `gemmi` and expected properties of the functions/twin laws.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -143,8 +143,45 @@ class DataSet(pd.DataFrame):
             return self.loc[~self.CENTRIC]
         return self.loc[~self.label_centrics().CENTRIC]
 
+    @property
+    def reindexing_ops(self):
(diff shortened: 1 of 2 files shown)


kmdalton (approved):
I find nothing objectionable here. This is just a simple wrapper for a helpful `gemmi` function. Some of the tests may be superfluous, but they're not hurting anything. I'd say go ahead and merge.

PR #187, opened 2022-09-29
Add return_edges argument to DataSet.assign_resolution_bins()

### LINKED ISSUE #184: rs.DataSet.assign_resolution_bins ought to return bin edges
The current bin labels assume formatting that takes up too much space. It would be better if this method just returned bin edges which users can treat however they want.

JBGreisman:
Fixes #184 by adding a `return_edges` argument to `DataSet.assign_resolution_bins()`. This PR adds that flexibility to the API, but otherwise does not change the default behavior of the function to preserve backwards compatibility.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -15,6 +15,7 @@ from reciprocalspaceship.decorators import (
 from reciprocalspaceship.dtypes.base import MTZDtype, MTZInt32Dtype
 from reciprocalspaceship.utils import (
     apply_to_hkl,
+    assign_with_binedges,
     bin_by_percentile,
     canonicalize_phases,
     compute_structurefactor_multiplicity,
@@ -891,8 +892,9 @@ class DataSet(pd.DataFrame):
 
         Parameters
         ----------
-        bins : int
-            Number of bins (default: 20)
+        bins : int, list, or np.ndarray
+            Number of bins or bin edges to use when assigning resolution bins. If
+            bin edges are provided, they must be monotonic (default: 20)
         inplace : bool
             Whether to add the column in place or return a copy (default: False)
         return_labels : bool
@@ -911,8 +913,16 @@ class DataSet(pd.DataFrame):
         """
         dHKL = self.compute_dHKL()["dHKL"]
 
-        assignments, edges = bin_by_percentile(dHKL, bins=bins, ascending=False)
-        self["bin"] = DataSeries(assignments, dtype="I", index=self.index)
+        if isinstance(bins, int):
+            assignments, edges = bin_by_percentile(dHKL, bins=bins, ascending=False)
+            self.loc[:, "bin"] = DataSeries(assignments, dtype="I", index=self.index)
+        else:
+            mask = (dHKL >= min(bins)) & (dHKL <= max(bins))
+            assignments = assign_with_binedges(dHKL[mask], bin_edges=bins)
+            result = self
+            result.loc[mask, "bin"] = assignments
+            if inplace:
+                self._update_inplace(result)
 
         # Package return values
         result = [self]
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -920,19 +920,14 @@ class DataSet(pd.DataFrame):
 
         if isinstance(bins, int):
             assignments, edges = bin_by_percentile(dHKL, bins=bins, ascending=False)
-            self.loc[:, "bin"] = DataSeries(assignments, dtype="I", index=self.index)
         else:
+            # Use bin edges for assignments and drop reflections outside of range
             mask = (dHKL >= min(bins)) & (dHKL <= max(bins))
             assignments = assign_with_binedges(dHKL[mask], bin_edges=bins)
             edges = np.array(bins)
+            self._update_inplace(self.loc[mask])
 
-            # Drop reflections outside of resolution range
-            result = self.loc[mask].copy()
-            result["bin"] = DataSeries(assignments, dtype="I", index=result.index)
-            if inplace:
-                self._update_inplace(result)
-            else:
-                self = result
+        self.loc[:, "bin"] = DataSeries(assignments, dtype="I", index=self.index)
 
         # Package return values
         result = [self]
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -879,7 +879,12 @@ class DataSet(pd.DataFrame):
 
     @inplace
     def assign_resolution_bins(
-        self, bins=20, inplace=False, return_labels=True, return_edges=False
+        self,
+        bins=20,
+        inplace=False,
+        return_labels=True,
+        format_str=".2f",
+        return_edges=False,
     ):
         """
         Assign reflections in DataSet to resolution bins.
@@ -893,8 +898,11 @@ class DataSet(pd.DataFrame):
         return_labels : bool
             Whether to return a list of labels corresponding to the edges
             of each resolution bin (default: True)
+        format_str : str
+            Format string for constructing bin labels
         return_edges : bool
-            Whether to return bin edges that define the resolution bin boundaries
+            Whether to return bin edges that define the resolution bin boundaries.
+            The bin edges are returned as a 1-dimensional array with `bins + 1` entries
             (default: False)
 
         Returns
@@ -903,12 +911,16 @@ class DataSet(pd.DataFrame):
         """
         dHKL = self.compute_dHKL()["dHKL"]
 
-        assignments, labels, edges = bin_by_percentile(dHKL, bins=bins, ascending=False)
+        assignments, edges = bin_by_percentile(dHKL, bins=bins, ascending=False)
         self["bin"] = DataSeries(assignments, dtype="I", index=self.index)
 
         # Package return values
         result = [self]
         if return_labels:
+            labels = [
+                f"{e1:{format_str}} - {e2:{format_str}}"
+                for e1, e2 in zip(edges[0:-1], edges[1:])
+            ]
             result.append(labels)
         if return_edges:
             result.append(edges)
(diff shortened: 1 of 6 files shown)


reciprocalspaceship/utils/binning.py
```diff
@@ -58,7 +58,10 @@ def assign_with_binedges(data, bin_edges, right_inclusive=True):
 
 
 def bin_by_percentile(
```
kmdalton, line 60: If anyone has used this function in their code, this will change the default behavior. I'd move that if we're going to change the default behavior we may as well just get rid of the labels entirely at the `utils` level and migrate the format string stuff to the top level API. This has the side effect of letting the formatting be user-facing via the intended API.

reciprocalspaceship/dataset.py
```diff
@@ -878,33 +878,44 @@ def compute_multiplicity(self, inplace=False, include_centering=True):
...
+            Number of bins (default: 20)
         inplace : bool
-            Whether to add the column in place or return a copy
+            Whether to add the column in place or return a copy (default: False)
         return_labels : bool
             Whether to return a list of labels corresponding to the edges
-            of each resolution bin
+            of each resolution bin (default: True)
+        return_edges : bool
+            Whether to return bin edges that define the resolution bin boundaries
```
kmdalton, line 897: I always find it hard to infer what "bin edges" means. For instance, is it an bins X 2 array or array with  bins + 1 entries? This needs to be clearer.

kmdalton (changes requested):
I have suggested a slight refactor. I haven't gone over the tests very carefully yet. I want to wait for your feedback on my suggestion.

reciprocalspaceship/dataset.py
JBGreisman: I'll specify that it's a 1d array with bins+1 entries

reciprocalspaceship/utils/binning.py
JBGreisman: yep. I think it's fine to change the API in `utils` because it's not intended to be user-facing. I agree with you that the formatting should be handled in `DataSet.assign_resolution_bins()` instead of the `utils` function

JBGreisman:
I made the requested refactors to `bin_by_percentile()` and the corresponding changes to `DataSet.assign_resolution_bins()` to preserve the user-facing API

kmdalton (changes requested):

reciprocalspaceship/dataset.py
```diff
@@ -878,33 +878,56 @@ def compute_multiplicity(self, inplace=False, include_centering=True):
...
+            Format string for constructing bin labels
+        return_edges : bool
+            Whether to return bin edges that define the resolution bin boundaries.
+            The bin edges are returned as a 1-dimensional array with `bins + 1` entries
+            (default: False)
 
         Returns
         -------
-        (DataSet, list) or DataSet
+        (DataSet, list, ndarray) or (DataSet, list) or DataSet
```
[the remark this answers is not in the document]
JBGreisman: yep -- I agree with both points and will update the order and add the fourth output

reciprocalspaceship/dataset.py
```diff
@@ -878,33 +878,56 @@ def compute_multiplicity(self, inplace=False, include_centering=True):
...
         self["bin"] = DataSeries(assignments, dtype="I", index=self.index)
 
+        # Package return values
+        result = [self]
         if return_labels:
-            return self, labels
-        else:
+            labels = [
+                f"{e1:{format_str}} - {e2:{format_str}}"
+                for e1, e2 in zip(edges[0:-1], edges[1:])
```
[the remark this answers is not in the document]
JBGreisman: sure -- no strong opinion from me.

reciprocalspaceship/stats/completeness.py
```diff
@@ -91,9 +91,9 @@ def compute_completeness(
     if dmin:
         dataset = dataset.loc[dHKL >= dmin]
         dHKL = dHKL[dHKL > dmin]
-    assignments, labels, binedges = bin_by_percentile(
-        dHKL, bins=bins, ascending=False, return_edges=True
-    )
+    assignments, binedges = bin_by_percentile(dHKL, bins=bins, ascending=False)
```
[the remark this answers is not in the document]
JBGreisman: done

JBGreisman:
@kmdalton I think this is good to go now, but let me know if you have any other thoughts

reciprocalspaceship/dataset.py
```diff
@@ -878,33 +878,56 @@ def compute_multiplicity(self, inplace=False, include_centering=True):
         return self
 
     @inplace
-    def assign_resolution_bins(self, bins=20, inplace=False, return_labels=True):
+    def assign_resolution_bins(
+        self,
+        bins=20,
```
kmdalton, line 883: I think you should be able to supply an array of bin edges.

reciprocalspaceship/stats/completeness.py
```diff
@@ -3,7 +3,7 @@
 
 import reciprocalspaceship as rs
 from reciprocalspaceship.utils.asu import in_asu
-from reciprocalspaceship.utils.binning import assign_with_binedges, bin_by_percentile
+from reciprocalspaceship.utils.binning import assign_with_binedges
```
kmdalton, line 6: I think you can accomplish the same thing you are using `assign_with_binedges` to do by using `groupby(['H', 'K', 'L']).first()`. I'm not sure if that would be better or not. Just thinking out loud.

kmdalton (changes requested):
I think `assign_resolution_bins` should take an array of bin edges. i think refls that fall outside the bin range should probably get dropped by default, but i am open to suggestions.

reciprocalspaceship/stats/completeness.py
JBGreisman: I don't quite follow -- the `groupby()` call would return the first instances of each Miller index, but `assign_with_binedges` returns which resolution bin each reflection gets assigned to. In this case, the goal is to use the resolution bins determined for the provided `dataset` to assign the reflections in the 100% complete instance. 

If we make your above suggestion (to have `assign_resolution_bins()`) take bin edges, then we could refactor this to use that function instead of `utils`.

kmdalton: okay we should probably do it via an `assign_resolution_bins` which accepts bin edges.

JBGreisman:
> i think refls that fall outside the bin range should probably get dropped by default, but i am open to suggestions.

This is different than the current behavior in `rs.utils.assign_with_binedges()`, but could be implemented in `DataSet.assign_resolution_bins()`by pre-dropping any reflections before the call. If I do not do the pre-processing, the behavior will be to raise a `ValueError` that says `"This function assumes 'bin_edges' contain every entry in 'data'" `.

I hesitate to auto-drop reflections from the DataSet, but I think that could be acceptable as long we add a corresponding remark in the docstring

reciprocalspaceship/dataset.py
JBGreisman: New commits support this

JBGreisman:
I updated this PR to allow `DataSet.assign_resolution_bins()` to support making assignments based on explicit bin edges. I made the following decisions regarding API:

- Reflections outside of provided bin edges are dropped
- Bin edges can be provided in ascending or descending order, and assignments will depend on that order

reciprocalspaceship/dataset.py
```diff
@@ -878,33 +879,75 @@ def compute_multiplicity(self, inplace=False, include_centering=True):
...
+        if isinstance(bins, int):
+            assignments, edges = bin_by_percentile(dHKL, bins=bins, ascending=False)
+            self.loc[:, "bin"] = DataSeries(assignments, dtype="I", index=self.index)
+        else:
+            mask = (dHKL >= min(bins)) & (dHKL <= max(bins))
+            assignments = assign_with_binedges(dHKL[mask], bin_edges=bins)
+            edges = np.array(bins)
+
+            # Drop reflections outside of resolution range
+            result = self.loc[mask].copy()
```
kmdalton, line 930: ideally you would not make a copy in the case of `inplace=True`.

tests/test_dataset_binning.py
```diff
@@ -0,0 +1,127 @@
...
+    Test DataSet.assign_resolution_bins() assignments depend on whether bin edges
+    are provided in ascending or descending order
+    """
+
+    expected, edges = data_fmodel.assign_resolution_bins(
+        2, return_labels=False, return_edges=True
+    )
+
+    # Flip bin edges -- all assignments should differ
+    if reverse:
```
kmdalton, line 91: can this be made to test for correctness? right now it only tests to see the result is not incorrect in one particular way.

reciprocalspaceship/dataset.py
```diff
@@ -878,33 +879,75 @@ def compute_multiplicity(self, inplace=False, include_centering=True):
...
+        return_labels=True,
+        format_str=".2f",
+        return_edges=False,
+    ):
         """
         Assign reflections in DataSet to resolution bins.
 
+        Notes
+        -----
+        - If bin edges are provided, any reflections outside of the specified range
```
kmdalton, line 895: good.

reciprocalspaceship/dataset.py
JBGreisman: While I agree in theory, I don't think it's possible to do so here without hitting a `SettingWithCopyWarning`. Let me know if you can think of a way to do so here -- I've tried all permutations I could think of.

tests/test_dataset_binning.py
JBGreisman: Sure -- this tests correctness against the assignments made using `assign_resolution_bins(2)`, but I can also check against `dHKL`, explicitly

JBGreisman: yep -- just updated this to test explicitly against `(1 - bin)` in the reverse case

reciprocalspaceship/dataset.py
JBGreisman: nvm -- I _think_ I got it

JBGreisman: resolved with new commits

kmdalton (approved):
looks good to me. great stuff!

PR #202, opened 2023-02-07
Fix documentation build by lowering jinja2 version

JBGreisman:
This is a quick fix to the documentation build. There is a current issue with sphinx due to a newer version of Jinja2 (https://github.com/readthedocs/readthedocs.org/issues/9037#issuecomment-1077818554).

After this fix, I'll look into making a more future-proof documentation build -- might require updating to using a more recent Python version when building documentation.


JBGreisman:
This fix corrects documentation builds locally -- our current CI doesn't explicitly test the build of the documentation, and it would be a good idea to add such a check in the future.

PR #207, opened 2023-02-20
Add GitHub Action for testing documentation build during pull requests

JBGreisman:
This PR adds a GitHub Action workflow that builds the static documentation site using sphinx. This addresses a lingering issue that the documentation build is only attempted upon merging into the `main` branch. This way we can catch issues before merging. 

This workflow is equivalent to the documentation build, but any changes are not added to the `gh-pages` branch and the site itself is not updated.

### DIFF
--- a/.github/workflows/testdocumentation.yml
+++ b/.github/workflows/testdocumentation.yml
@@ -1,0 +1,36 @@
+name: Test Documentation
+
+on:
+  pull_request:
+    branches:
+      - main
+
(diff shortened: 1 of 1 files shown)


JBGreisman:
This problem was raised by #202, where merging a new PR led to a new version of jinja2 being used for building the documentation. This was not caught by our build tests of the PR, and was only realized when the site build failed.

Issue #206, opened 2023-02-15
API reference website display


DHekstra:
https://rs-station.github.io/reciprocalspaceship/api/index.html somehow shows all of the text subsumed in the warning box.

![image](https://user-images.githubusercontent.com/35602603/219144090-eab695c6-62a8-499f-9c52-b67712d9b3fc.png)

PR #208, opened 2023-02-21
Upgrade sphinx version by removing sphinx_panels dependency

JBGreisman:
The documentation used a deprecated sphinx extension called `sphinx_panels`. This update switches to the updated version called `sphinx_design`. This should make it so that a later version of sphinx can be used for building documentation, which resolves the jinja2 version issue from #202 and as a side effect also resolves the issue identified in #206.

This worked locally without the need for setting an explicit minimal sphinx version in `setup.py`, but it might behave differently in the CI.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,7 +40,7 @@ extensions = [
     "sphinx_rtd_theme",
     "nbsphinx",
     "sphinx.ext.linkcode",
-    "sphinx_panels",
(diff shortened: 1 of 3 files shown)


JBGreisman:
Based on the documentation test, the changes here did indeed boost the sphinx version from `3.5.3 --> 5.3.0`

Issue #213, opened 2023-05-15
Return keys of dictionary in crystfel.py


DHekstra:
It would be helpful to optionally return the keys of the dictionary that `_parse_stream` in `io/crystfel.py` produces when calling `rs.read_crystfel`. My use case is working with stream files produced by a new crystfel data processing pipeline at LCLS that contains data from multiple fixed-target crystals. The crystal ID and per-crystal frame numbers are not encoded in the BATCH record produced by `rs.read_crystfel` but are recoverable from the keys. There's no need to parse the keys any further--just to return `list(d.keys())`.

JBGreisman:
Is this going to be a new, general feature of the crystfel stream file output? If it is a very customized output for the LCLS processing pipeline (and will not make it more broadly into crystfel), I don't know if it makes sense to support. 

For very personal/custom use cases, I think it may be best to put together your own implementation rather than modifying the interface here.

DHekstra:
My use case is, of course, specific, but the point is general: `_parse_stream()` already returns a dictionary with keys and values. I am simply asking for the option for `read_crystfel()` to also return _that_ dictionary, or at least a list of its keys.

JBGreisman:
got it -- I'll think over whether that will make sense. In the meantime, the desired functionality can be obtained this way (this is not officially supported, and is subject to change...):

```python
from reciprocalspaceship.io.crystfel import _parse_stream
d, cell = _parse_stream(streamfile)
```

DHekstra:
That can work. In that case, could we turn lines 238-268 in `crystfel.py` into another function that can be called, e.g. `dict_to_dataset()` or so? That would allow me to write clean code that can both get the DataSet and the dictionary.

JBGreisman:
The `StreamLoader` class provided by #216 to speed up the parsing of stream files may provide the info you want in a cleaner way. I'm not exactly sure what data you want, but you will be able to instantiate a `StreamLoader` object using something like this:
```python
from reciprocalspaceship.io.crystfel import StreamLoader
loader = StreamLoader(streamfile)              # use to find metadata you seek
ds = loader.to_dataset(spacegroup=spacegroup)  # also get a DataSet
ds.set_index(["H", "K", "L"], inplace=True)
```
This isn't yet in the main codebase, but you can always test it out using the `faster_stream` branch of the repo

kmdalton:
[...]

PR #216, opened 2023-06-29
Faster parser for CrystFEL stream files

kmdalton:
I wrote a faster version of the stream file parser. It mostly just uses more numpy and less pandas while vectorizing operations within a crystal block.

### DIFF
--- a/reciprocalspaceship/io/crystfel.py
+++ b/reciprocalspaceship/io/crystfel.py
@@ -1,34 +1,64 @@
+from typing import Optional, Union
+
+import gemmi
 import numpy as np
 import pandas as pd
-from typing import Union,Optional
 
 import reciprocalspaceship as rs
 from reciprocalspaceship import DataSet
-import gemmi
+from reciprocalspaceship.decorators import cellify, spacegroupify
 from reciprocalspaceship.utils import angle_between
-from reciprocalspaceship.decorators import cellify,spacegroupify
 
-#See Rupp Table 5-2
+# See Rupp Table 5-2
 _cell_constraints = {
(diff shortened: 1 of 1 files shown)


DHekstra:
Does this/could this also address https://github.com/rs-station/reciprocalspaceship/issues/213 ?

kmdalton:
@DHekstra ,  i don't know if the information you want is available in the stream file. i would  need to see an example.

JBGreisman:
This looks pretty good to me, but I'll do another look through before approving/merging.

kmdalton:
There are a couple of issues that still need to be addressed
 1) Doeke provided me some more recent stream files, and they seem to break some of my formatting assumptions. 
 2) The current version doesn't respect that some files have the photon energy measured per frame. My code just uses the photon energy in the header for every crystal. This is a problem for XFEL data where the peak energy changes from shot to shot. 

I'm going to hold off until I have a bit more time to look through it.

DHekstra:
@kmdalton, what is the status of this? I'm revisiting some stream files next week, and could test drive your parser.

DHekstra:
that's alright. I have a slow but functional parser.

JBGreisman:
Just checking in here -- is there anything left to do on this PR or should it be merged into main?

kmdalton:
I discovered this version in was incompatible with some stream files provided by @DHekstra. I haven't been able to diagnose the issue yet. I would recommend we hold off on merging.

JBGreisman:
Does this need to be merged anymore? Or will this be superseded by #260 ?

PR #235, opened 2023-12-13
clip I for Sigma calc

DHekstra:
When using `scale_merged_intensities` to perform French-Wilson scaling in the `anisotropic` mode, the local Sigma is calculated by a local weighted average of observed intensities. As a consequence, Sigma can be negative. This leads to NaNs in the calculation of the logpdf used to calculate the posterior expected intensity. To remedy this, I propose clipping the observed intensities to be positive in (and only in) the calculation of Sigma in `mean_intensity_by_miller_index(I, H, bandwidth)`.

### DIFF
--- a/reciprocalspaceship/algorithms/scale_merged_intensities.py
+++ b/reciprocalspaceship/algorithms/scale_merged_intensities.py
@@ -95,6 +95,7 @@ def _french_wilson_posterior_quad(Iobs, SigIobs, Sigma, centric, npoints=100):
     return mean, np.sqrt(variance), mean_F, np.sqrt(variance_F)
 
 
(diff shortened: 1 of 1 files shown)


kmdalton:
I'm not certain that fw should really be applied to data with resolution shells that have negative average intensity. Silently clipping the values is not a good solution if it leads to the user unknowingly processing such data. If you want to merge a change like this, it needs to print a warning or better yet be implemented as an optional flag which is disabled by default.

DHekstra:
That makes sense. In my single test, Sigma was negative for ~0.01% of reflections. I'll think some more about this when time permits.

PR #237, opened 2023-12-17
Support clipping Sigma to avoid negative values in French-Wilson method

DHekstra:
I added a flag permitting clipping of Iobs to positive values for the purpose of calculating Sigma during anisotropic FW calculation. The option is off by default.

### DIFF
--- a/reciprocalspaceship/algorithms/scale_merged_intensities.py
+++ b/reciprocalspaceship/algorithms/scale_merged_intensities.py
@@ -95,8 +95,12 @@ def _french_wilson_posterior_quad(Iobs, SigIobs, Sigma, centric, npoints=100):
     return mean, np.sqrt(variance), mean_F, np.sqrt(variance_F)
 
 
+<<<<<<< HEAD
 
 def mean_intensity_by_miller_index(I, H, bandwidth, clip_neg_Iobs):
+=======
+def mean_intensity_by_miller_index(I, H, bandwidth):
+>>>>>>> d25d7da19615acf5c6694af20766c9a7cbeb80b2
     """
     Use a gaussian kernel smoother to compute mean intensities as a function of miller index.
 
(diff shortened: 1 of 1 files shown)


kmdalton:
Wouldn't it make more sense to clip the returned values rather than the input intensities? This version of the code adds small errors to _all_ miller indices rather than just the ones which would be assigned negative average intensities.

kmdalton:
Sigma

DHekstra:
cool. makes sense.

DHekstra:
OK, this seems to do the trick. Clipping sigma to a user-specified small positive value (too small results in numerical warnings).

DHekstra:
```
import numpy as np

ds=rs.read_mtz('pipeline/data/mtzs_origin/111.mtz')
ds_out=rs.algorithms.scale_merged_intensities(\
        ds, \
        intensity_key="IMEAN", \
        sigma_key="SIGIMEAN", \
        output_columns=["IP","SIGIP","FP","SIGFP"], \
        dropna=True, \
        inplace=False, \
        mean_intensity_method='anisotropic', \
        bw=2.0,
        clip_neg_Sigma=False,
        eps=0.0,
    )
```
produces
```
print(np.sum(np.isnan(ds_out["FP"].to_numpy())))

22238
```
ds_out=rs.algorithms.scale_merged_intensities(\
        ds, \
        intensity_key="IMEAN", \
        sigma_key="SIGIMEAN", \
        output_columns=["IP","SIGIP","FP","SIGFP"], \
        dropna=True, \
        inplace=False, \
        mean_intensity_method='anisotropic', \
        bw=2.0,
        clip_neg_Sigma=True,
        eps=0.1,
    )
```
produces
```
print(np.sum(np.isnan(ds_out["FP"].to_numpy())))
0
```

kmdalton:
I think it's better to just have a single new parameter, minimum_sigma=-np.inf

DHekstra:
Let me know how this looks. I switched to using a single flag `minimum_sigma` as @kmdalton suggested, with a default value of `-np.inf` (no minimum). Because some of the ~440 datasets also yielded some negative Sigma with the `isotropic` method, I moved the clipping into `scale_merged_intensities()` itself. I am applying the clipping _before_ multiplication by the multiplicity as that strikes me as most appropriate. Let me know how this looks.

kmdalton:
I am happy with this change. There might be a better parameter name (mea culpa), but otherwise this is the least invasive patch I can think of. Let's ask @JBGreisman for comments before merging, but I'm plenty satisfied.

JBGreisman (approved):
This looks fine to me -- I agree that it makes sense to have the default behavior as unchanged relative to before, and the proposed call to `np.clip` seems fine for getting this functionality.

PR #247, opened 2024-03-11
support **kwargs in gemmification decoratored fn

kmdalton:
This PR modifies the `@spacegroupify` and `@cellify` decorators so as to support decorating functions which take arbitrary keyword arguments (`**kwargs`). Specifically, it alters the usage of the `BoundArguments` instance in the decorated function call. 

`f(**bargs.arguments)` becomes `f(*bargs.args, **bargs.kwargs)`


JBGreisman (approved):
This looks good -- makes sense to update these decorators to fully support arbitrary keyword arguments

PR #258, opened 2024-06-17
enable numpy 2.0 support

kmdalton:
This PR enables `rs` to run on `numpy` version 2.0 which is also supported by the latest `pandas` release. The major changes required were
 - Squeeze extra dim from inverse returned by `np.unique`
 - `np.find_common_type` was removed in version 2.0, but pandas provides an equivalent method, `from pandas.core.dtypes.cast import np_find_common_type`
 - Remove `np.NaN` which is now exclusively `np.nan`
 - `np.array` no longer supports `np.array(list, copy=False`. So I added some control flow which relies on `np.asarray`.


JBGreisman (approved):
This all looks good to me. Thanks for looking into the API changes with v2.0

PR #259, opened 2024-06-17
Is numeric na

kmdalton:
This PR speeds up dataset creation for large arrays. In working on a new stream file parser, I realized that converting numpy float32 and int32  to MTZDtypes  could be very time-consuming for large arrays. After some digging, I found out the culprit was a cython function provided by pandas, `pandas._libs.missing.is_numeric_na` which makes a mask for missing values in `rs` 
 - ['floating.py`](https://github.com/rs-station/reciprocalspaceship/blob/2ef8ae44b3d3d456cf6ba2efdcd751c65c04330c/reciprocalspaceship/dtypes/floating.py#L163)
 - ['integer.py'](https://github.com/rs-station/reciprocalspaceship/blob/2ef8ae44b3d3d456cf6ba2efdcd751c65c04330c/reciprocalspaceship/dtypes/integer.py#L205)

In the case that the input is an int32 or float32 numpy array, this is wholly unnecessary, and it is much faster to use `np.isnan` to accomplish the same task. This PR just wraps is_numeric_na and adds some control flow to accomplish that. It falls back to the Cython version whenever the input is not an int32 or float32 ndarray. This is a very conservative choice, and more circumstances could probably be included in the control flow down the line.

### DIFF
--- a/reciprocalspaceship/dtypes/floating.py
+++ b/reciprocalspaceship/dtypes/floating.py
@@ -34,7 +34,6 @@ from __future__ import annotations
 
 import numpy as np
 from pandas._libs import lib
-from pandas._libs import missing as libmissing
 from pandas._typing import ArrayLike, DtypeObj
 from pandas.util._decorators import cache_readonly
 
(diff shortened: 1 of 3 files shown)


kmdalton:
This should wait to merge until after #258

JBGreisman (commented):
Now that #258 is merged in, I think that some of these changes are unnecessary in this PR. Do you mind updating/rebasing so that the relevant changes aren't overwritten here?

kmdalton:
Okay, I cleaned up this PR. It should be good now. I will merge after the CI runs.

PR #260, opened 2024-06-19
Parallel Stream File Parsing with ray.io

kmdalton:
This PR adds support for faster stream file parsing which is parallelized using the [ray](ray.io). I did not add ray as a dependency for users, so the code falls back to serial python when it is not available.

### DIFF
--- a/reciprocalspaceship/io/crystfel.py
+++ b/reciprocalspaceship/io/crystfel.py
@@ -1,13 +1,11 @@
 import mmap
 import re
-from os.path import getsize
 from typing import Union
 
 import gemmi
 import numpy as np
-import pandas as pd
 
-from reciprocalspaceship import DataSeries, DataSet, concat
+from reciprocalspaceship import DataSeries, DataSet
 from reciprocalspaceship.utils import angle_between, eV2Angstroms
 
 # See Rupp Table 5-2
--- a/reciprocalspaceship/io/crystfel.py
+++ b/reciprocalspaceship/io/crystfel.py
@@ -525,9 +525,6 @@ def read_crystfel(
     --------
     rs.DataSet
     """
-    from time import time
-
-    start = time()
     if not streamfile.endswith(".stream"):
         raise ValueError("Stream file should end with .stream")
 
--- a/reciprocalspaceship/io/crystfel.py
+++ b/reciprocalspaceship/io/crystfel.py
@@ -2,8 +2,8 @@ import re
 from mmap import mmap
 from os.path import getsize
 from typing import Union
-import gemmi
 
+import gemmi
 import numpy as np
 import pandas as pd
 
@@ -192,14 +192,14 @@ class StreamLoader(object):
 
     def calculate_average_unit_cell(self) -> gemmi.UnitCell:
         """
-        Compute the average of all cell parameters across the file. 
+        Compute the average of all cell parameters across the file.
         """
         regex = re.compile(rb"Cell parameters .+\n")
         with open(self.filename, "r+") as f:
             memfile = mmap(f.fileno(), 0)
             lines = regex.findall(memfile)
-        cell = np.loadtxt(lines, usecols=[2,3,4,6,7,8], dtype='float32').mean(0)
-        cell[:3] *= 10.
+        cell = np.loadtxt(lines, usecols=[2, 3, 4, 6, 7, 8], dtype="float32").mean(0)
+        cell[:3] *= 10.0
 
         header = self.extract_file_header()
         lattice_type = None
--- a/reciprocalspaceship/io/crystfel.py
+++ b/reciprocalspaceship/io/crystfel.py
@@ -567,7 +567,6 @@ def read_crystfel(
     cell = loader.extract_target_unit_cell()
 
     batch = 0
-    batch_array = []
     ds = []
 
     for chunk in loader.read_crystfel(
--- a/reciprocalspaceship/io/crystfel.py
+++ b/reciprocalspaceship/io/crystfel.py
@@ -2,6 +2,7 @@ import re
 from mmap import mmap
 from os.path import getsize
 from typing import Union
+import gemmi
 
 import numpy as np
 import pandas as pd
@@ -182,6 +183,28 @@ class StreamLoader(object):
                     cell = [None] * 6
                 value = float(line.split()[2])
                 cell[idx] = value
+            if line.startswith("lattice_type ="):
+                lattice_type = line.split()[-1]
+
+        if lattice_type is not None:
+            cell = _cell_constraints[lattice_type](cell)
+        return cell
+
+    def calculate_average_unit_cell(self) -> gemmi.UnitCell:
+        """
+        Compute the average of all cell parameters across the file. 
+        """
+        regex = re.compile(rb"Cell parameters .+\n")
+        with open(self.filename, "r+") as f:
+            memfile = mmap(f.fileno(), 0)
+            lines = regex.findall(memfile)
+        cell = np.loadtxt(lines, usecols=[2,3,4,6,7,8], dtype='float32').mean(0)
+        cell[:3] *= 10.
+
+        header = self.extract_file_header()
+        lattice_type = None
+
+        for line in header.split("\n"):
             if line.startswith("lattice_type ="):
                 lattice_type = line.split()[-1]
 
--- a/reciprocalspaceship/io/crystfel.py
+++ b/reciprocalspaceship/io/crystfel.py
@@ -1,5 +1,6 @@
 import mmap
 import re
+from importlib.util import find_spec
 from typing import Union
 
 import gemmi
@@ -7,8 +8,6 @@ import numpy as np
 
 from reciprocalspaceship import DataSeries, DataSet
 from reciprocalspaceship.utils import angle_between, eV2Angstroms
-
-from importlib.util import find_spec
 
 # See Rupp Table 5-2
 _cell_constraints = {
@@ -77,8 +76,8 @@ class StreamLoader(object):
         "L": 2,
         "I": 3,
         "SigI": 4,
-        "peak" : 5,
-        "background" : 6,
+        "peak": 5,
+        "background": 6,
         "XDET": 7,
         "YDET": 8,
         "s1x": 9,
@@ -268,9 +267,9 @@ class StreamLoader(object):
             A list of columns to include in the peak list numpy arrays.
             A list of possible column names is stored as stream_loader.available_column_names.
         use_ray : bool(optional)
-            Whether or not to use ray for parallelization. 
+            Whether or not to use ray for parallelization.
         num_cpus : int (optional)
-            The number of cpus for ray to use. 
+            The number of cpus for ray to use.
         ray_kwargs : optional
             Additional keyword arguments to pass to [ray.init](https://docs.ray.io/en/latest/ray-core/api/doc/ray.init.html#ray.init).
 
@@ -287,11 +286,12 @@ class StreamLoader(object):
 
         # Check whether ray is available
         if use_ray:
-            if find_spec('ray') is None:
+            if find_spec("ray") is None:
                 use_ray = False
                 import warnings
+
                 message = (
-                    "ray (https://www.ray.io/) is not available..." 
+                    "ray (https://www.ray.io/) is not available..."
                     "Falling back to serial stream file parser."
                 )
                 warnings.warn(message, ImportWarning)
@@ -304,6 +304,7 @@ class StreamLoader(object):
             )
             if use_ray:
                 import ray
+
                 ray.init(num_cpus=num_cpus, **ray_kwargs)
 
                 @ray.remote
@@ -463,7 +464,13 @@ class StreamLoader(object):
 
 
 def read_crystfel(
-    streamfile: str, spacegroup=None, encoding="utf-8", columns=None, parallel=True, num_cpus=None, **ray_kwargs
+    streamfile: str,
+    spacegroup=None,
+    encoding="utf-8",
+    columns=None,
+    parallel=True,
+    num_cpus=None,
+    **ray_kwargs,
 ) -> DataSet:
     """
     Initialize attributes and populate the DataSet object with data from a CrystFEL stream with indexed reflections.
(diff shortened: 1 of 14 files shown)


JBGreisman:
This all looks good to me, but I'd like to play with it a bit more before merging. I think I agree with making `ray` an optional dependency, but I don't think I like adding it to `tests_require` -- seems like a hacky solution.

What are your thoughts on us adding an explicit `parallel_require=["ray"]`, that is added to `[dev]` and maybe a new `[parallel]` pip option that only adds on the `ray` extra requirement?

kmdalton:
I'm happy to defer to your preferences regarding requirements. I don't have strong feelings as long as we make it easy for users to figure out how to get parallelism. I am not yet familiar enough with `ray` to know how nicely it plays with other packages. So far it seems very promising.

kmdalton:
@marinegor , would you be willing to test out this branch for us?

marinegor:
@kmdalton sure, I can have a look! what kind of testing are you thinking, could you elaborate? I imagine you want to make sure that your parser produces same results as the previous one, right?

kmdalton:
Thank you! I don't have access to a lot of stream files, and I have noticed there can be some differences in the metadata between files. Mostly I want to make sure I'm not missing anything which will break the parser for some edge cases. Additionally, I would hope you could let us know 
 - Whether it is reasonably  that it requires an additional dependency, `ray`, in order to run in parallel
 - Whether the `read_crystfel` and the `StreamLoader` class are adequately documented
 - Whether the parser runs in parallel as expected
 - Whether the parser has reasonable memory usage relative to the original implementation
 - Whether it generates the expected output

DHekstra:
$ `conda install -c conda-forge "ray-default"`

fails for me like this in a fresh conda environment in which I (tried and maybe failed) to use your careless install script:

```
Channels:
 - conda-forge
 - defaults
Platform: linux-64
Collecting package metadata (repodata.json): done
Solving environment: failed

LibMambaUnsatisfiableError: Encountered problems while solving:
  - nothing provides _python_rc needed by python-3.12.0rc3-rc3_hab00c5b_1_cpython

Could not solve for environment specs
The following packages are incompatible
├─ python 3.12**  is installable with the potential options
│  ├─ python [3.12.0|3.12.1|3.12.2|3.12.3|3.12.4], which can be installed;
│  ├─ python [3.12.0|3.12.1|3.12.2|3.12.3|3.12.4] would require
│  │  └─ python_abi 3.12.* *_cp312, which can be installed;
│  └─ python 3.12.0rc3 would require
│     └─ _python_rc, which does not exist (perhaps a missing channel);
└─ ray-default is not installable because there are no viable options
   ├─ ray-default [1.10.0|1.11.0|...|2.0.0] would require
   │  ├─ python >=3.7,<3.8.0a0 , which conflicts with any installable versions previously reported;
   │  └─ python_abi 3.7.* *_cp37m, which conflicts with any installable versions previously reported;
   ├─ ray-default [1.10.0|1.11.0|...|2.9.3] would require
   │  ├─ python >=3.8,<3.9.0a0 , which conflicts with any installable versions previously reported;
   │  └─ python_abi 3.8.* *_cp38, which conflicts with any installable versions previously reported;
   ├─ ray-default [1.10.0|1.11.0|...|2.9.3] would require
   │  ├─ python >=3.9,<3.10.0a0 , which conflicts with any installable versions previously reported;
   │  └─ python_abi 3.9.* *_cp39, which conflicts with any installable versions previously reported;
   ├─ ray-default [1.13.0|2.0.0|...|2.9.3] would require
   │  ├─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
   │  └─ python_abi 3.10.* *_cp310, which conflicts with any installable versions previously reported;
   ├─ ray-default [1.5.0|1.5.1|1.5.2|1.6.0] would require
   │  ├─ python >=3.6,<3.7.0a0 , which conflicts with any installable versions previously reported;
   │  └─ python_abi 3.6.* *_cp36m, which conflicts with any installable versions previously reported;
   ├─ ray-default [2.10.0|2.11.0|...|2.9.3] would require
   │  ├─ python >=3.11,<3.12.0a0 , which conflicts with any installable versions previously reported;
   │  └─ python_abi 3.11.* *_cp311, which conflicts with any installable versions previously reported;
   ├─ ray-default 2.8.0 would require
   │  └─ ray-core 2.8.0 py38h1702d6c_1, which does not exist (perhaps a missing channel);
   ├─ ray-default [1.6.0|1.9.2|2.0.1] would require
   │  └─ python >=3.7,<3.8.0a0 , which conflicts with any installable versions previously reported;
   ├─ ray-default [1.6.0|1.9.2|2.0.1|2.3.0|2.6.3] would require
   │  └─ python >=3.8,<3.9.0a0 , which conflicts with any installable versions previously reported;
   ├─ ray-default [1.6.0|1.9.2|2.0.1|2.3.0|2.6.3] would require
   │  └─ python >=3.9,<3.10.0a0 , which conflicts with any installable versions previously reported;
   ├─ ray-default [2.0.1|2.3.0|2.6.3] would require
   │  └─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
   └─ ray-default 2.6.3 would require
      └─ python >=3.11,<3.12.0a0 , which conflicts with any installable versions previously reported.
```

DHekstra:
This also fails: 

```
(careless-13)[dhekstra@holy8a24301 reciprocalspaceship]$ pip install -U "ray"
ERROR: Could not find a version that satisfies the requirement ray (from versions: none)
ERROR: No matching distribution found for ray

(careless-13)[dhekstra@holy8a24301 reciprocalspaceship]$ pip install -U "ray[default]"
ERROR: Could not find a version that satisfies the requirement ray[default] (from versions: none)
ERROR: No matching distribution found for ray[default]

(careless-13)[dhekstra@holy8a24301 reciprocalspaceship]$ pip install ray
ERROR: Could not find a version that satisfies the requirement ray (from versions: none)
ERROR: No matching distribution found for ray
```

kmdalton:
@DHekstra , I think some of your packages (careless for sure) do not have python 3.12 support which is confusing the package solver. You should use python 3.11 for now.

marinegor (changes requested):
Hey @kmdalton , sorry for the long delay -- hope it'll go faster from here.
Great PR, thanks for implementing it! It certainly needs a re-work, and ti's really cool that you did it :)

Short changes summary (detailed below):
 - allow running without `ray` even if it's installed
 - run `ray` in context manager to gracefully shutdown if interrupted
 - `r+` reading rights break when stream files are accessible readonly

And that's it! Really great work, thanks again for inviting me :)

---

Detailed answers to your questions:

> 0. I don't have access to a lot of stream files, and I have noticed there can be some differences in the metadata between files.

If you want smoke tests, I'd ask Filippe di Maia from CXIDB (https://cxidb.org/contact.html) for contacts -- I'm sure he could find you few dozens of various streams for testing.
Or you can parse CXIDB for links to stream files -- I wrote a parser with BeautifulSoup, it's doable.

> 1. Whether it is reasonably that it requires an additional dependency, ray, in order to run in parallel

A: it's perfectly reasonable, as long as it's possible to fall back to non-ray option even if ray is installed.
I left a comment about that, but basically, being able to run without `ray` even if it's installed, is necessary.


> 2. Whether the read_crystfel and the StreamLoader class are adequately documented

I'd add important ray keywords explicitly, and examples with how to limit CPU and RAM usage.


> 3. Whether the parser runs in parallel as expected

It seems that yes, and even `num_cpus` influences that as well.


1. Whether the parser has reasonable memory usage relative to the original implementation
It probably heavily depends on the stream size, and I'd add some rule-of-thumb rules based on that. For example, in case of CXIDB ID106's stream, previous implementation was much faster for me, since the ray cluster's overhead is too big.
I can easily imagine that ray can give performance boosts when dealing with e.g. >2Gb streams or something, depending on IO speed, but it's probably better to document this.


5. Whether it generates the expected output

That I can't say, but it can easily be tested -- I'd take 3-4 stream files from CXIDB (e.g. check [ID106](https://cxidb.org/data/106/C1_Zaf_P21_stream.tar.gz), [ID223](https://cxidb.org/id-223.html) -- combined they have 5 stream files).
You can make up syntetic datasets (one crystal, two crystals, five crystals), and test if they give comparable results.

kmdalton:
@marinegor , thanks so much for this feedback. It'll take me some time to work through it, but this is super helpful stuff. 🙏

reciprocalspaceship/io/crystfel.py
```diff
@@ -1,270 +1,553 @@
...
+        if lattice_type is not None:
+            cell = _cell_constraints[lattice_type](cell)
+        return cell
+
+    def calculate_average_unit_cell(self) -> gemmi.UnitCell:
+        """
+        Compute the average of all cell parameters across the file.
+        """
+        regex = re.compile(rb"Cell parameters .+\n")
+        with open(self.filename, "r+") as f:
```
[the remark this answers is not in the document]
kmdalton: addressed by c11e99f2657ef8e24f2f351287cc5a58a9aa5e68

reciprocalspaceship/io/crystfel.py
```diff
@@ -1,270 +1,553 @@
...
+            memfile = mmap(f.fileno(), 0)
+            lines = regex.findall(memfile)
+        cell = np.loadtxt(lines, usecols=[2, 3, 4, 6, 7, 8], dtype="float32").mean(0)
+        cell[:3] *= 10.0
```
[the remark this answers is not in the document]
kmdalton: Addressed by a1e44a367263cb4759707e479d31f696c34fdc3d

reciprocalspaceship/io/crystfel.py
```diff
@@ -1,270 +1,553 @@
...
+            peak_list_columns = [self.peak_list_columns[s] for s in peak_list_columns]
+
+        with open(self.filename, "r+") as f:
+            memfile = mmap(f.fileno(), 0)
+            beginnings_and_ends = zip(
+                self.block_regex_bytes["chunk_begin"].finditer(memfile),
+                self.block_regex_bytes["chunk_end"].finditer(memfile),
+            )
+            try:
+                import ray
```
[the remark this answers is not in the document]
kmdalton: i added an option to disable ray, and i changed the code to check before trying to import ray (84b83ad07b9024ccc3bf8f42f46ad3dfda2ef35c). if we end up with more functions that rely on ray, we can set the environment variable on module import. however, i want to leave this as is right now in case someone decides to do something like `!pip install ray` from an active IPython terminal or Jupyter notebook.

reciprocalspaceship/io/crystfel.py
```diff
@@ -1,270 +1,553 @@
...
-                crystal_idx += 1
-                is_crystal = True
-                continue
-
-    return answ_crystals, rv_cell_param
-
-
-def read_crystfel(streamfile: str, spacegroup=None) -> DataSet:
+                results = ray.get(result_ids)
+                ray.shutdown()
```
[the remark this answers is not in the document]
kmdalton: I added something like you requested in a03780aa033fdf833f52c36a46015b981cefc3ae. I'm not sure if it ticks all your boxes, would you mind having a look? @marinegor

reciprocalspaceship/io/crystfel.py
```diff
@@ -1,270 +1,608 @@
...
-
-    return dataset
+    for k in columns:
+        mtz_dtypes[k] = mtz_dtypes.get(k, "R")
+
+    loader = StreamLoader(streamfile, encoding=encoding)
+    cell = loader.extract_target_unit_cell()
+
+    batch = 0
+    batch_array = []
```
marinegor, line 570: unused variable

reciprocalspaceship/io/crystfel.py
```diff
@@ -1,270 +1,608 @@
...
+    ray_kwargs : optional
+        Additional keyword arguments to pass to [ray.init](https://docs.ray.io/en/latest/ray-core/api/doc/ray.init.html#ray.init).
 
     Returns
     --------
     rs.DataSet
     """
+    from time import time
 
+    start = time()
```
marinegor, line 530: unused variable

reciprocalspaceship/io/crystfel.py
```diff
@@ -1,270 +1,608 @@
...
+import re
+from contextlib import contextmanager
+from importlib.util import find_spec
+from typing import Union
 
+import gemmi
+import numpy as np
 
-def _parse_stream(filename: str) -> dict:
+from reciprocalspaceship import DataSeries, DataSet, concat
```
marinegor, line 10: unused DataSeries

marinegor (approved):
@kmdalton thanks for the swift reply, I think it's good to go now! Nice changes, and on CXIDB's ID106 it indeed gives the same results:

image

which is amazing, given how many rounding errors could've been there :)

I added few nitpicking comments that popped up in my IDE, and that's it!

reciprocalspaceship/io/crystfel.py
kmdalton: removed in f1770ac

reciprocalspaceship/io/crystfel.py
kmdalton: removed in 33b57be

reciprocalspaceship/io/crystfel.py
kmdalton: ec6ad73

reciprocalspaceship/io/crystfel.py
```diff
@@ -1,270 +1,553 @@
...
+    encoding : str
+        The type of byte-encoding (optional, 'utf-8').
+    columns : list (optional)
+        Optionally specify the columns of the output by a list of strings.
+        The default list is:
+            [ "H", "K", "L", "I", "SigI", "BATCH", "s1x", "s1y", "s1z", "ewald_offset",
+            "angular_ewald_offset", "XDET", "YDET" ]
+        See `rs.io.crystfel.StreamLoader().available_column_names` for a list of available column names.
+    ray_kwargs : optional
+        Additional keyword arguments to pass to [ray.init](https://docs.ray.io/en/latest/ray-core/api/doc/ray.init.html#ray.init).
```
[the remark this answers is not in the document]
kmdalton: done

JBGreisman (approved):
This all looks good to me. Feel free to merge if/when you're happy with it.

PR #261, opened 2024-06-24
Restore Numpy v1 Support

kmdalton:
The goal of this pr is to
 - restore numpy v1 support which was broken by #258 
 - modify the CI to build against numpy v1 and v2


JBGreisman (approved):
This looks good to me -- seems good to maintain compatibility wherever possible, and I like the expansion of the test suite around numpy versions.

PR #265, opened 2024-08-15
Add support for range indexed data in `DataSet.write_mtz()` and `DataSet.to_gemmi()`

### LINKED ISSUE #255: RS requires set_index(["H","K","L"]) to create a multi_index using rs.DataSet
Problem: Derek was trying to construct an `rs.DataSet` object from NumPy arrays (h, k, l, wave, etc.) for unmerged data. The following, 

```
 ds = rs.DataSet({"H":h, "K":k, "L":l, "WAVE": wave, "dH":dh, "dK":dk, "dL":dl,
               "NPIX": npix_col, "BATCH": batch, "SigI": sigI_vals, "I": I_vals, "X":x, "Y":y},
               cell=args.ucell, spacegroup=sg_num, merged=False)
    ds["PARTIAL"] = True
    ds.write_mtz(args.mtz)
```

this throws the following error.

```
In [1]: ds.write_mtz("wow.mtz")

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[1], line 1
----> 1 ds.write_mtz("wow.mtz")

File /global/cfs/cdirs/lcls/dermen/postcori/xtal/conda_base/lib/python3.8/site-packages/reciprocalspaceship/dataset.py:612, in DataSet.write_mtz(self, mtzfile, skip_problem_mtztypes, project_name, crystal_name, dataset_name)
    586 """
    587 Write DataSet to MTZ file.
    588 
   (...)
    608     Dataset name to assign to MTZ file
    609 """
    610 from reciprocalspaceship import io
--> 612 return io.write_mtz(
    613     self,
    614     mtzfile,
    615     skip_problem_mtztypes,
    616     project_name,
    617     crystal_name,
    618     dataset_name,
    619 )

File /global/cfs/cdirs/lcls/dermen/postcori/xtal/conda_base/lib/python3.8/site-packages/reciprocalspaceship/io/mtz.py:225, in write_mtz(dataset, mtzfile, skip_problem_mtztypes, project_name, crystal_name, dataset_name)
    191 def write_mtz(
    192     dataset,
    193     mtzfile,
   (...)
    197     dataset_name,
    198 ):
    199     """
    200     Write an MTZ reflection file from the reflection data in a DataSet.
    201 
   (...)
    223         Dataset name to assign to MTZ file
    224     """
--> 225     mtz = to_gemmi(
    226         dataset, skip_problem_mtztypes, project_name, crystal_name, dataset_name
    227     )
    228     mtz.write_to_file(mtzfile)
    229     return

File /global/cfs/cdirs/lcls/dermen/postcori/xtal/conda_base/lib/python3.8/site-packages/reciprocalspaceship/io/mtz.py:151, in to_gemmi(dataset, skip_problem_mtztypes, project_name, crystal_name, dataset_name)
    149         continue
    150     else:
--> 151         raise ValueError(
    152             f"column {c} of type {cseries.dtype} cannot be written to an MTZ file. "
    153             f"To skip columns without explicit MTZ dtypes, set skip_problem_mtztypes=True"
    154         )
    155 mtz.set_data(temp[columns].to_numpy(dtype="float32"))
    157 # Handle Unmerged data

ValueError: column index of type int64 cannot be written to an MTZ file. To skip columns without explicit MTZ dtypes, set skip_problem_mtztypes=True

```

Changing the last line by adding `infer_mtz_datatypes().set_index(["H","K","L"], drop=True)` solves the problem.

```
ds.infer_mtz_dtypes().set_index(["H","K","L"], drop=True).write_mtz(args.mtz)
```

I'll leave it to @dermen and @kmdalton to comment further.

JBGreisman:
PR fixes #255. 

Currently, DataSet objects with a pd.RangeIndex() (for example, due to a call to `DataSet.reset_index()`) end up causing trouble due to an extra call to `reset_index()` within `rs.io.mtz.to_gemmi()`. To resolve this issue, this PR uses the `range_indexed` decorator to ensure that the relevant functions can assume the calling `DataSet` object is always range indexed. This simplifies the control flow for these functions.

As part of this PR, I also modified `range_indexed` to handle use cases where the decorated function returns None or non-DataSet types.

### DIFF
--- a/reciprocalspaceship/decorators.py
+++ b/reciprocalspaceship/decorators.py
@@ -3,6 +3,7 @@ from inspect import signature
 
 import gemmi
 import numpy as np
+import reciprocalspaceship as rs
 
 
 def inplace(f):
@@ -49,6 +50,9 @@ def range_indexed(f):
         if result is None:
             ds._index_from_names(names, inplace=True)
             return
+        if not isinstance(result, rs.DataSet):
+            ds._index_from_names(names, inplace=True)
+            return result
         result = result._index_from_names(names, inplace=True)
         ds = ds._index_from_names(names, inplace=True)
         return result.__finalize__(ds)
--- a/reciprocalspaceship/decorators.py
+++ b/reciprocalspaceship/decorators.py
@@ -47,14 +47,12 @@ def range_indexed(f):
         names = ds.index.names
         ds = ds._index_from_names([None], inplace=True)
         result = f(ds, *args, **kwargs)
+        ds = ds._index_from_names(names, inplace=True)
         if result is None:
-            ds._index_from_names(names, inplace=True)
             return
         if not isinstance(result, rs.DataSet):
-            ds._index_from_names(names, inplace=True)
             return result
         result = result._index_from_names(names, inplace=True)
-        ds = ds._index_from_names(names, inplace=True)
         return result.__finalize__(ds)
 
     return wrapped
--- a/reciprocalspaceship/decorators.py
+++ b/reciprocalspaceship/decorators.py
@@ -46,6 +46,9 @@ def range_indexed(f):
         names = ds.index.names
         ds = ds._index_from_names([None], inplace=True)
         result = f(ds, *args, **kwargs)
+        if result is None:
+            ds._index_from_names(names, inplace=True)
+            return
         result = result._index_from_names(names, inplace=True)
         ds = ds._index_from_names(names, inplace=True)
         return result.__finalize__(ds)
--- a/reciprocalspaceship/decorators.py
+++ b/reciprocalspaceship/decorators.py
@@ -3,6 +3,7 @@ from inspect import signature
 
 import gemmi
 import numpy as np
+
 import reciprocalspaceship as rs
 
 
--- a/reciprocalspaceship/decorators.py
+++ b/reciprocalspaceship/decorators.py
@@ -49,12 +49,10 @@ def range_indexed(f):
         ds = ds._index_from_names([None], inplace=True)
         result = f(ds, *args, **kwargs)
         ds = ds._index_from_names(names, inplace=True)
-        if result is None:
-            return
-        if not isinstance(result, rs.DataSet):
-            return result
-        result = result._index_from_names(names, inplace=True)
-        return result.__finalize__(ds)
+        if isinstance(result, rs.DataSet):
+            result = result._index_from_names(names, inplace=True)
+            result = result.__finalize__(ds)
+        return result
 
     return wrapped
 
--- a/reciprocalspaceship/io/mtz.py
+++ b/reciprocalspaceship/io/mtz.py
@@ -142,7 +142,6 @@ def to_gemmi(
             columns.append(c)
         # Special case for CENTRIC and PARTIAL flags
         elif cseries.dtype.name == "bool" and c in ["CENTRIC", "PARTIAL"]:
-            dataset[c] = dataset[c].astype("MTZInt")
             mtz.add_column(label=c, type="I")
             columns.append(c)
         elif skip_problem_mtztypes:
--- a/reciprocalspaceship/io/mtz.py
+++ b/reciprocalspaceship/io/mtz.py
@@ -133,16 +133,16 @@ def to_gemmi(
     mtz.datasets[0].dataset_name = dataset_name
 
     # Construct data for Mtz object
-    temp = dataset.reset_index()
+    # GH#255: DataSet is provided using the range_indexed decorator
     columns = []
-    for c in temp.columns:
-        cseries = temp[c]
+    for c in dataset.columns:
+        cseries = dataset[c]
         if isinstance(cseries.dtype, MTZDtype):
             mtz.add_column(label=c, type=cseries.dtype.mtztype)
             columns.append(c)
         # Special case for CENTRIC and PARTIAL flags
         elif cseries.dtype.name == "bool" and c in ["CENTRIC", "PARTIAL"]:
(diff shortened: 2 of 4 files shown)


reciprocalspaceship/decorators.py
```diff
@@ -46,8 +48,12 @@ def wrapped(ds, *args, **kwargs):
         names = ds.index.names
         ds = ds._index_from_names([None], inplace=True)
         result = f(ds, *args, **kwargs)
-        result = result._index_from_names(names, inplace=True)
         ds = ds._index_from_names(names, inplace=True)
+        if result is None:
```
kmdalton, line 52: I would prefer to use the following pattern, while acknowledging it is a purely aesthetic preference. 

```python
if isinstance(result, rs.DataSet):
    result = result._index_from_names(names, inplace=True)
    result =  result.__finalize__(ds)
return result
```

reciprocalspaceship/io/mtz.py
```diff
@@ -133,16 +133,16 @@ def to_gemmi(
...
-        cseries = temp[c]
+    for c in dataset.columns:
+        cseries = dataset[c]
         if isinstance(cseries.dtype, MTZDtype):
             mtz.add_column(label=c, type=cseries.dtype.mtztype)
             columns.append(c)
         # Special case for CENTRIC and PARTIAL flags
         elif cseries.dtype.name == "bool" and c in ["CENTRIC", "PARTIAL"]:
-            temp[c] = temp[c].astype("MTZInt")
+            dataset[c] = dataset[c].astype("MTZInt")
```
kmdalton, line 145: this line changes the dtype of the passed dataset's 'CENTRIC' column. can we avoid this side effect? one way would be to convert it back to bools after saving, but maybe there is a cleaner way? 

```
[ins] In [23]: ds = rs.read_mtz("unmerged.mtz")[['I', 'SIGI']].label_centrics()

[ins] In [24]: ds.dtypes
Out[24]:
I          Intensity
SIGI          Stddev
CENTRIC         bool
dtype: object

[ins] In [25]: ds.write_mtz('test.mtz')

[ins] In [26]: ds.dtypes
Out[26]:
I          Intensity
SIGI          Stddev
CENTRIC       MTZInt
PARTIAL         bool
dtype: object
```

kmdalton: A new "PARTIAL" column is also created. I'd prefer it if this wasn't added. It's clear from the docstring there's something special about PARTIAL and CENTRIC, but I didn't expect my dataset to be modified by calling this method.

tests/io/test_mtz.py
```diff
@@ -223,3 +223,32 @@ def test_write_mtz_names(IOtest_mtz, project_name, crystal_name, dataset_name):
 
     # Clean up
     temp.close()
+
+
+@pytest.mark.parametrize("range_indexed", [True, False])
+def test_write_mtz_rangeindexed(data_merged, range_indexed):
```
kmdalton, line 229: Such a simple, helpful, well-written test. You love to see it.

tests/io/test_mtz.py
```diff
@@ -223,3 +223,32 @@ def test_write_mtz_names(IOtest_mtz, project_name, crystal_name, dataset_name):
...
+    """
+    if range_indexed:
+        data_merged = data_merged.reset_index()
+
+    with tempfile.NamedTemporaryFile(suffix=".mtz") as temp:
+        data_merged.write_mtz(temp.name)
+        assert exists(temp.name)
+
+
+@pytest.mark.parametrize("range_indexed", [True, False])
```
kmdalton, line 241: good stuff

kmdalton (changes requested):
I think we need to do something about the side-effects on "PARTIAL" and "CENTRIC" columns. Otherwise, this looks like a fantastic PR.

reciprocalspaceship/io/mtz.py
JBGreisman: That's a fair point, but if I'm not mistaken, these are pre-existing side-effects that are unrelated to this change, right? The CENTRIC/PARTIAL columns need to be converted to an MTZ compatible column type to be written out. The "PARTIAL" in your example gets added because it's an unmerged dataset -- that's also a requirement for M/ISYM handling in the MTZ file format. 

I agree that it would be nice to avoid any changes to the DataSet upon writing -- the simplest way to do that would be to copy the dataset prior to changes, but I'd rather avoid unnecessary memory/computation. Do you think these changes are best handled here? Or should we file an additional Issue and handle in the future?

reciprocalspaceship/decorators.py
JBGreisman: agreed. Cleaner. I'll update.

reciprocalspaceship/io/mtz.py
JBGreisman: Ah, this may have been introduced as a side effect by this PR. Previously, `reset_index()` had been called without `inplace=True`, whereas it is now called inplace due to the `range_indexed` decorator. As such, in the past these operations were done on a copy of the calling dataset. 

As such, I think the simplest fix is to copy the calling dataset. It feels suboptimal, but it does preserve the previous behavior.

JBGreisman: Ok I think I fixed this in the latest set of commits

kmdalton (approved):
this looks squeaky clean now. thanks! one thought for the future is whether we can set one column of Gemmi MTZ data at a time so you don't have to call `to_numpy` on the whole dataset.

JBGreisman:
Yep -- that might be a worthwhile enhancement if we ever want to look into performance-related improvements. 

I'll merge this in

PR #266, opened 2024-08-16
Update `rs.DataSet.reset_index()` call signature to matches `pandas >v1.5`

JBGreisman:
There was an update to the pandas API in v1.5 that changed the `pd.DataFrame.reset_index()` call signature. This PR updates the call signature of `rs.DataSet.reset_index()` and adds tests to confirm the consistency of the call signature between:
- `rs.DataSet.reset_index()` and `pd.DataFrame.reset_index()`
- `rs.DataSet.set_index()` and `pd.DataFrame.set_index()`

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -258,7 +258,14 @@ class DataSet(pd.DataFrame):
         )
 
     def reset_index(
-        self, level=None, drop=False, inplace=False, col_level=0, col_fill=""
+        self,
+        level=None,
+        drop=False,
+        inplace=False,
+        col_level=0,
+        col_fill="",
+        allow_duplicates=lib.no_default,
+        names=None,
     ):
         """
         Reset the index or a specific level of a MultiIndex.
@@ -281,6 +288,12 @@ class DataSet(pd.DataFrame):
         col_fill : object
             If the columns have multiple levels, determines how the other
             levels are named. If None then the index name is repeated.
+        allow_duplicates : bool
+            Allow duplicate column labels to be created.
+        names : int, str, tuple, list
+            Using the given string, rename the DataSet column which contains the
+            index data. If the DataSet has a MultiIndex, this has to be a list or
+            tuple with length equal to the number of levels.
 
         Returns
         -------
@@ -317,6 +330,8 @@ class DataSet(pd.DataFrame):
                 inplace=inplace,
                 col_level=col_level,
                 col_fill=col_fill,
+                allow_duplicates=allow_duplicates,
+                names=names,
             )
             _handle_cached_dtypes(self, columns, drop)
             return
@@ -327,6 +342,8 @@ class DataSet(pd.DataFrame):
                 inplace=inplace,
                 col_level=col_level,
                 col_fill=col_fill,
+                allow_duplicates=allow_duplicates,
+                names=names,
             )
             dataset._index_dtypes = dataset._index_dtypes.copy()
             dataset = _handle_cached_dtypes(dataset, columns, drop)
--- a/tests/test_dataset_signatures.py
+++ b/tests/test_dataset_signatures.py
@@ -1,0 +1,41 @@
+from inspect import signature
+
+import pandas as pd
+import pytest
+from pandas.testing import assert_frame_equal
+
+
+def test_reset_index_signature(dataset_hkl):
+    """
+    Test call signature of rs.DataSet.reset_index() matches call signature of
+    pd.DataFrame.reset_index() using default parameters
+    """
+    df = pd.DataFrame(dataset_hkl)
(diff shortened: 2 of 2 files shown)


tests/test_dataset_signatures.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Minimal example from GH#223
+    """
+    result = rs.DataSeries(range(10)).reset_index()
+    expected = pd.Series(range(10)).reset_index()
+    expected = rs.DataSet(expected)
+    assert_frame_equal(result, expected)
+
+
+def test_reset_index_signature(dataset_hkl):
```
kmdalton, line 20: This looks like a good way to test whether pandas changed. Maybe we should do this test for more functions.

reciprocalspaceship/dataset.py
```diff
@@ -258,7 +258,14 @@ def set_index(
         )
 
     def reset_index(
```
kmdalton, line 260: my question here is why don't use a pattern like

```python
@wraps(pd.DataFrame.reset_index)
def reset_index(self, level, *args, **kwargs):
    ...
    super().reset_index(level, *args, **kwargs)
```
Wouldn't that be more robust to Pandas API changes?

tests/test_dataset_signatures.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Test call signature of rs.DataSet.reset_index() matches call signature of
+    pd.DataFrame.reset_index() using default parameters
+    """
+    df = pd.DataFrame(dataset_hkl)
+    sig = signature(pd.DataFrame.reset_index)
+    bsig = sig.bind(df)
+    bsig.apply_defaults()
+
+    expected = df.reset_index(*bsig.args[1:], **bsig.kwargs)
```
kmdalton, line 30: You test the results, but is there any way to directly test the equivalency of the signatures?

kmdalton (approved):
I'm going to approve this, because it seems sensible and fixes the bug, but I wonder if it'd be better to use a more robust approach based on `functools.wraps` and `*args, **kwargs`.

tests/test_dataset_signatures.py
JBGreisman: The only other methods we directly overload and call `super()` on use the *args, **kwargs paradigm to overload, so I didn't bother adding the tests. I'm open to adding such methods, but it didn't seem like such a test would actually be informative (because they directly pass through arguments).

reciprocalspaceship/dataset.py
JBGreisman: I was considering making that sort of change here. I've never figured out how to do it in a way that makes the docstrings nice. We do use that for `to_pickle()` and a few other overwritten methods, but I figured I'd keep the same form here because these are very commonly used operations and I like having the explicit docstring that refers to `DataSet` objects, etc. I'm open to changing this if you feel strongly.

tests/test_dataset_signatures.py
kmdalton: cool. sounds like a non-issue. still a good trick to keep in mind.

reciprocalspaceship/dataset.py
kmdalton: no i don't have strong feelings, it just seemed like a missed opportunity. i also don't know how to avoid mangling the docstrings.

tests/test_dataset_signatures.py
JBGreisman: I could go through the `bsig.parameters.keys()` and make sure they match in the same order. I thought of this test as "good enough" because it would flag any changes to argument names. I like the idea that in the future we could add new arguments/default values if we need to, which is why I didn't quite want to directly assess signatures.

kmdalton: okay. sounds fine for now.

JBGreisman:
I'm open to the idea of the `*args`/`**kwargs` paradigm and we do use it elsewhere. Personally, I find these to be very frequently used methods, so I enjoy having the easy introspection of call signature and correct docstrings (that don't mention `pandas` specifics). That was the main rationale for keeping it this way. Definitely open to reconsidering if you feel otherwise though.

kmdalton:
> I'm open to the idea of the `*args`/`**kwargs` paradigm and we do use it elsewhere. Personally, I find these to be very frequently used methods, so I enjoy having the easy introspection of call signature and correct docstrings (that don't mention `pandas` specifics). That was the main rationale for keeping it this way. Definitely open to reconsidering if you feel otherwise though.

I think there are good reasons not to use `*args`/ `**kwargs` in this case. It's quite an important method. I was just trying to safeguard against changes in Pandas breaking the CI. I think maybe the CI should break in this case. We should hear about it whenever their call signature changes. I'm good with this.

Issue #272, opened 2024-10-28
`pkg_resources` no longer in `python3.12+`


tjlane:
I'm trying to expand the set of python build versions for `meteor`, but `reciprocalspaceship` can't support anything past `3.11` because it's still using `pkg_resources`, luckily in just one place:

https://github.com/rs-station/reciprocalspaceship/blob/ceae60e293bfdb3e969d0e3e2b53fa3a2b9e34f9/reciprocalspaceship/__init__.py#L1-L9

This issue from another project explains the situation:
https://github.com/mu-editor/mu/issues/2485

In general I'd be pro bumping the build python versions. They are pretty old at this point. Thoughts?

kmdalton:
I'm in favor of this as well, but I fear that's not the only thing we need to do in order to support 3.12. Do we know for certain that's the only thing we need to change? I think we probably also have to migrate to `pyproject.toml`, but I could be wrong.

kmdalton:
fixed by #273

PR #273, opened 2024-10-28
add python 3.12 support

kmdalton:
Attempting to address #272 by updating package versioning to be independent of `pkg_resources`.


kmdalton:
Also fixing Codecov #278

PR #276, opened 2024-10-28
pyproject.toml

tjlane:
Re #273 -- seems this is not necessary, but it's easy if you want it... and you do want it.


tjlane:
@kmdalton hmm i guess my PR into PR doesn't trigger the CI tests? And as an outsider, I can't request a reviewer specifically. But: check it out.

Issue #277, opened 2024-11-04
bug in `is_isomorphous` cell comparison


tjlane:
The method documentation says:
https://github.com/rs-station/reciprocalspaceship/blob/ceae60e293bfdb3e969d0e3e2b53fa3a2b9e34f9/reciprocalspaceship/dataset.py#L1191-L1192

But the code looks for an absolute change:
https://github.com/rs-station/reciprocalspaceship/blob/ceae60e293bfdb3e969d0e3e2b53fa3a2b9e34f9/reciprocalspaceship/dataset.py#L1217

By default this will return `False` if the cells differ by more than 5 Å or 5 degrees in any case.

I'd suggest we change this to make it relative, as the documentation indicates.

kmdalton:
Can confirm. This is a bug. I like your proposal.

kmdalton:
what's a reasonable default for the `cell_threshold`?  i think 0.5% is in the right ballpark. @tjlane

kmdalton:
fixed in #279

PR #279, opened 2024-11-04
Fix is_isomorphous

kmdalton:
As discussed in #277 , DataSet.is_isomorphous currently does not do what is indicated in its docstring. This change makes the behavior in line with the expectation that `cell_threshold` is a percent difference.

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -605,10 +605,10 @@ def test_is_isomorphous(data_unmerged, data_fmodel, sg1, sg2, cell1, cell2):
 
 @pytest.mark.parametrize("threshold", [5.0, 1.0, 0.5, 0.1])
 @pytest.mark.parametrize("isomorphous", [True, False])
-@pytest.mark.parametrize("bigger_cell", ['self', 'other']) #Which cell is bigger
+@pytest.mark.parametrize("bigger_cell", ["self", "other"])  # Which cell is bigger
 def test_is_isomorphous_threshold(threshold, isomorphous, bigger_cell):
     """
-    Test that DataSet.is_isorphous(self, other, cell_threshold) method's 
+    Test that DataSet.is_isorphous(self, other, cell_threshold) method's
     cell_threshold operates on percent difference.
     """
     epsilon = 1e-12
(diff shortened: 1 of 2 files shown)


tjlane (approved):
Bug is fixed.

I'd like to see a (regression) test for this method added before merging, to ensure it *stays* fixed.

kmdalton:
as requested, @tjlane

tests/test_dataset.py
```diff
@@ -603,6 +603,40 @@ def test_is_isomorphous(data_unmerged, data_fmodel, sg1, sg2, cell1, cell2):
...
+        if isomorphous:  # shrink
+            factor = factor * (1.0 - epsilon)
+        else:  # grow
+            factor = factor * (1.0 + epsilon)
+    else:  # cell_test < cell
+        factor = (200.0 - threshold) / (200.0 + threshold)
+        if isomorphous:  # grow
+            factor = factor * (1.0 + epsilon)
+        else:  # shrink
+            factor = factor * (1.0 - epsilon)
```
tjlane, line 631: This test is smart... maybe too smart? It had to read it quite a few times before seeing it clearly.

Generally I like dumb stuff. More like:

```python
if isomorphous:
    cell_change_factor = (100.0 + threshold) / 100.0 - epsilon
else:
    cell_change_factor = (100.0 + threshold) / 100.0 + epsilon

if sign:
    cell_test = cell * cell_change_factor
else:
    cell_test = cell / cell_change_factor
```

or something similar.

tests/test_dataset.py
```diff
@@ -603,6 +603,40 @@ def test_is_isomorphous(data_unmerged, data_fmodel, sg1, sg2, cell1, cell2):
             assert not result
 
 
+@pytest.mark.parametrize("threshold", [5.0, 1.0, 0.5, 0.1])
+@pytest.mark.parametrize("isomorphous", [True, False])
+@pytest.mark.parametrize("sign", [True, False])  # Which cell is bigger
```
tjlane, line 608: `sign` is a bit confusing, I anticipated something being positive or negative.

What about `@pytest.mark.parametrize("compare_to", ["bigger_cell", "smaller_cell"])`, then compare strings in what follows?

tjlane (commented):
Nice! Logic looks solid. A few comments in the interest of people who might read this code in the future.

Also: do we need a test for the spacegroup comparison bit?

tests/test_dataset.py
kmdalton: Okay I clarified it.

tests/test_dataset.py
kmdalton: Okay I think I made it dumber. Let me know.

kmdalton:
@tjlane , space groups are tested in  `test_is_isomorphous`. I don't think there is any point in doing it for this test too.

tjlane (approved):
love it

tjlane:
thanks for tackling this one @kmdalton !

PR #285, opened 2024-12-04
Deprecated `DataSet.to_reciprocalgrid()` should call `DataSet.to_reciprocal_grid()`

JBGreisman:
This is a very minor bookkeeping PR -- we currently have two function for converting a `DataSet` to a 3D numpy array representing the underlying data: `DataSet.to_reciprocalgrid()` and `DataSet.to_reciprocal_grid()`. The former will be deprecated in favor of the latter, in order to have consistency naming schemes in terms of underscores.

However, we currently have duplicate implementations for both functions. If we want to edit the implementations before we officially remove one of them, we will need to make changes in both places. This PR updates the deprecated code to call the newer version to avoid needing to maintain the code in two places.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -1585,24 +1585,6 @@ class DataSet(pd.DataFrame):
         warnings.simplefilter("always")
         warnings.warn(message, DeprecationWarning)
 
-        if dmin is not None:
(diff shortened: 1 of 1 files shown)


kmdalton (approved):
This looks good to me. Sorry I missed this in my feed.

PR #286, opened 2024-12-06
add hkls attribute rs.DataSet

kmdalton:
This PR makes a `.hkls` attribute for `rs.DataSet` instances. The method has a setter to assign new Miller indices. It retains backward compatibility with the `get_hkls` method. 

The use case here is for being able to set new miller indices using an expression like
```python
ds.hkls = hkls
```
where hkls can either be a dataset containing the keys 'H', 'K', and 'L' or it could be a numpy array. This saves the user having to reset and set the index.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -156,7 +156,7 @@ class DataSet(pd.DataFrame):
         return hkl
 
     def get_hkls(self):
-        """For backwards compatibility retain the get_hkls method in addition to the dataset.hkls attribute"""
+        """Get the Miller indices of the dataset."""
         return self.hkls
 
     @hkls.setter
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -53,7 +53,7 @@ class DataSet(pd.DataFrame):
     centrics : rs.DataSet
         Access only the centric reflections in this dataset
     hkls : ndarray, shape=(n_reflections, 3)
-        Miller indices in DataSet. 
+        Miller indices in DataSet.
     merged : bool
         Whether this is a merged dataset or unmerged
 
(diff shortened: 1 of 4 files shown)


JBGreisman:
I think the concept of this makes a lot of sense. I think having `DataSet.hkls` as an attribute is an intuitive accessor for this sort of info, and I agree with maintaining the `get_hkls()` method for compatibility. I still have to do a bit of a dive into this to see if there are any gotchas/corner cases that seem problematic. 

Some questions that will be easy to answer but I'm just writing here as part of my stream of consciousness: 
- Does this work correctly for both range-indexed and HKL-indexed DataSets (yes! this is explicitly tested)
- What about if there are extra columns in the index?
- What happens if the provided HKLs do not have the same number of rows as the DataSet? Do they get NaN-padded?

From what I can tell, this all seems fine, though

reciprocalspaceship/dataset.py
```diff
@@ -43,6 +43,24 @@ class DataSet(pd.DataFrame):
...
+        Access only the acentric reflections in this dataset
+    cell : gemmi.UnitCell
+        The unit cell
+    centrics : rs.DataSet
+        Access only the centric reflections in this dataset
+    hkls : ndarray, shape=(n_reflections, 3)
+        Miller indices in DataSet.
+    merged : bool
+        Whether this is a merged dataset or unmerged
+
```
JBGreisman, line 59: Should this extra space be here?

reciprocalspaceship/dataset.py
```diff
@@ -131,6 +149,38 @@ def merged(self):
...
 
+    @property
+    @range_indexed
+    def hkls(self):
+        """Miller indices"""
+        hkl = self[["H", "K", "L"]].to_numpy(dtype=np.int32)
+        return hkl
+
+    def get_hkls(self):
+        """For backwards compatibility retain the get_hkls method in addition to the dataset.hkls attribute"""
```
JBGreisman, line 160: I think this docstring should be more of a method description, rather than a rationale for keeping the method

tests/test_dataset.py
```diff
@@ -711,3 +711,44 @@ def test_select_mtzdtype_ValueError(data_merged, dtype):
...
+        input_ds = data_merged
+    else:
+        input_ds = data_unmerged
+
+    ds = input_ds.copy()
+
+    if range_index:
+        ds = ds.reset_index()
+
+    hmax = 20
```
JBGreisman, line 735: We have two pytest fixtures (`hkls` and `dataset_hkl`) that could probably support this test, rather than rolling your own. If the other two are insufficient, maybe we should modify them so that this can be used elsewhere as well?

JBGreisman (commented):
Just a few small comments -- overall, looks good

reciprocalspaceship/dataset.py
kmdalton: Removed.

reciprocalspaceship/dataset.py
kmdalton: Okay, I changed it. Let me know if you are okay with the update.

tests/test_dataset.py
kmdalton: i don't think these fixtures are appropriate, because the hkls in this case should be matched to the length of the dataset. it might make sense to use them to test the case where the dataset length and hkl length differ.

kmdalton: maybe we could match these hkls to those in data_merged and/or data_unmerged?

kmdalton: okay -- i changed the hkl fixtures to be matched to the data fixtures. let me know if you think this is a good solution.

tests/conftest.py
```diff
@@ -65,6 +39,44 @@ def data_unmerged():
...
+
+@pytest.fixture
+def dataset_hkl(data_merged):
+    """
+    Build DataSet for testing containing only Miller indices
+    """
+    H = data_merged.hkls
+    dataset = rs.DataSet({"H": H[:, 0], "K": H[:, 1], "L": H[:, 2]})
+    dataset.set_index(["H", "K", "L"], inplace=True)
+    return dataset
```
JBGreisman, line 58: Is `data_merged` 100% complete between [-5, 5] for all hkls? I found this to be a useful fixture for testing things like gemmi/sg operations in the past, so I hesitate to fully change it over to this, given that `data_merged` is an experimental HEWL dataset and not a complete synthetic one.

Sorry to be a pain, but I think this was better as it was before -- I worry a bit about collateral effects of changing these fixtures in this way.

JBGreisman (changes requested):
I think the `hkls` fixture should be changed back to be generated using numpy so that it is more clear what is being provided and avoid any collateral changes to other tests

tests/conftest.py
kmdalton: okay, it makes sense to be careful here. i reverted this change. i simplified the test code a little. that seems like a better solution than dataset-specific hkl fixtures.

JBGreisman (approved):
Perfect -- thanks! I think this looks good to me.

Issue #289, opened 2025-02-13
rs.mtzdump: have a `--complete` option?


DHekstra:
for MTZs containing more than ~7 columns, could we 

1. add a transpose step on the output of `mtz.describe()` so all stats are reported? 
2. the `mtz.dtypes()` already does something like that, but it would be great to get the dtypes for all columns. something like 

```
with pd.option_context('display.max_rows', None, 'display.max_columns', None):
    print(df)
```

here's an example of current output:
``` 
mtz.describe():

                I      SigI     BATCH  ...  xtal_142  xtal_146  xtal_148
count    332955.0  332955.0  332955.0  ...  332955.0  332955.0  332955.0
mean      112.716   718.411    802.06  ...       0.0       0.0       0.0
std       817.298   437.906   510.241  ...       0.0       0.0       0.0
min      -1265.39      30.9       1.0  ...       0.0       0.0       0.0
25%         -0.78     32.04     403.0  ...       0.0       0.0       0.0
50%         10.58    948.98     716.0  ...       0.0       0.0       0.0
75%         33.69    948.98    1186.0  ...       0.0       0.0       0.0
max    121984.141  10761.74    1943.0  ...       0.0       0.0       0.0

[8 rows x 67 columns]

mtz.dtypes:

I           Intensity
SigI           Stddev
BATCH           Batch
s1x           MTZReal
s1y           MTZReal
              ...    
xtal_137       MTZInt
xtal_142       MTZInt
xtal_146       MTZInt
xtal_148       MTZInt
PARTIAL          bool
Length: 68, dtype: object

```

kmdalton:
Can you share a pathological mtz file for development, @DHekstra ? Ideally truncate it to some small number of rows first. Thanks!

DHekstra:
yeah, give me a few hours (teaching). I can work with Luis on this. My current example does have an obscene number of rows indeed.

kmdalton:
resolved by #290

PR #290, opened 2025-02-13
transpose mtz description, add resolution info

kmdalton:
Address #289 by transposing the description in `rs.mtzdump`. It also calculates the resolution range of the data if the cell is set. 

The potentially contentious thing about this PR is that it changes the orientation of the description for `rs.mtzdump`. It would be easy to make this dynamic and/or user controlled, but my preference is to make it consistent across inputs. Since the transposed orientation works perfectly fine for any shape of input, I suggest we just use it exclusively. Happy to hear other takes on this.

### DIFF
--- a/reciprocalspaceship/commandline/mtzdump.py
+++ b/reciprocalspaceship/commandline/mtzdump.py
@@ -79,7 +79,7 @@ def summarize(mtz, precision):
             print(f"Resolution range:  {dHKL.max():.3f} - {dHKL.min():.3f} Å")
         with pd.option_context("display.max_rows", None):
             print(f"\nmtz.head():\n\n{mtz.head()}")
-            print(f"\nmtz.describe():\n\n{mtz.describe().T}")
+            print(f"\nmtz.describe().T:\n\n{mtz.describe().T}")
             print(f"\nmtz.dtypes:\n\n{mtz.dtypes}")
     return
 


kmdalton:
Here's a very wide mtz for testing:
[wide.mtz.zip](https://github.com/user-attachments/files/18785365/wide.mtz.zip)

DHekstra (approved):
Thank you. That does exactly what I was looking for!

kmdalton:
okay -- i will need to give @JBGreisman a chance to weight in

JBGreisman:
I think this is fine -- I agree that it is pesky to have the `mtz.describe()` abridged by the ellipses in some cases, and this seems like a perfectly legible form. My only suggestion would be to change the output line from `mtz.describe()` to `mtz.describe().T` to make it clear exactly what command would be needed to get the printed output. I'll put this in a code review.

reciprocalspaceship/commandline/mtzdump.py
```diff
@@ -74,9 +74,13 @@ def summarize(mtz, precision):
...
         )
-        print(f"\nmtz.head():\n\n{mtz.head()}")
-        print(f"\nmtz.describe():\n\n{mtz.describe()}")
-        print(f"\nmtz.dtypes:\n\n{mtz.dtypes}")
+        if mtz.cell is not None:
+            dHKL = mtz.compute_dHKL().dHKL
+            print(f"Resolution range:  {dHKL.max():.3f} - {dHKL.min():.3f} Å")
+        with pd.option_context("display.max_rows", None):
+            print(f"\nmtz.head():\n\n{mtz.head()}")
+            print(f"\nmtz.describe():\n\n{mtz.describe().T}")
```
JBGreisman, line 82: I think this should be changed to:
`print(f"\nmtz.describe().T:\n\n{mtz.describe().T}")`

to make it clear what line could be used to reproduce the printed stats.

JBGreisman (changes requested):
Looks good to me -- I put one small nitpicky comment, but I'm fine with the main idea.

JBGreisman (approved):
Perfect -- let's merge once the CI completes

PR #315, opened 2025-11-05
remove ray, add python 3.14 to build matrix

kmdalton:
Since transitioning from Ray to Joblib, we should be able to support python 3.14 now.


kmdalton:
apparently numpy 1.21.6 is extremely cursed on python 3.14. i am going to exclude the numpy 1, python 3.14 combo from the build matrix.

PR #322, opened 2025-12-03
Units cli

kmdalton:
Adding a very simple command line app to convert between ev and angstroms. This was prompted by having written the following line in a bash script:

```wavelength=`python -c "from reciprocalspaceship.utils import ev2angstroms;print(ev2angstroms($energy.))"```


JBGreisman (approved):
This looks good to me. I'll merge it in.

PR #323, opened 2025-12-03
propose fix for certain refl tables

kmdalton:
We have some reflection tables for which the `expt_id_map` isn't populated. I propose to fallback to bypassing this mapping, but I'm not sure if it's safe to do so.


kmdalton:
just to add a bit of context, our troublesome files were created by translating some old pickle files from dials to the msgpack format.  we used the following approach in python:

```python
from dials.array_family import flex
from dxtbx.model import ExperimentList
expt = ExperimentList.from_file("01.json", check_format=False)
refl = flex.reflection_table.from_file("01.pickle")
expt.as_file(f"output/01.expt")
refl.as_file(f"output/01.refl")
```

kmdalton:
okay i'll go ahead and merge this. thanks, @dermen

PR #326, opened 2025-12-23
Added additional msgpack data types for rs.io.read_dials_stills

LuisA92:
I encountered errors when using`rs.io.read_dials_stills(extra_cols=['bbox', 'entering'])`. The issue is resolved by adding additional data types for the problematic columns.


JBGreisman:
seems reasonable to me too. Once @dermen approves, looks like a good fix to improve coverage of different `dials` outputs

LuisA92:
For reference, the following columns from an `integrated.refl` work with the additional data types:

```python
extra_cols = [
'background.mean',
 'background.sum.value',
 'background.sum.variance',
 'bbox',
 'd',
 'entering',
 'flags',
 'id',
 'imageset_id',
 'intensity.prf.value',
 'intensity.prf.variance',
 'intensity.sum.value',
 'intensity.sum.variance',
 'lp',
 'miller_index',
 'num_pixels.background',
 'num_pixels.background_used',
 'num_pixels.foreground',
 'num_pixels.valid',
 'panel',
 'partial_id',
 'partiality',
 'profile.correlation',
 'qe',
 'refl_ids',
 's1',
 'xyzcal.mm',
 'xyzcal.px',
 'xyzobs.mm.value',
 'xyzobs.mm.variance',
 'xyzobs.px.value',
 'xyzobs.px.variance',
 'zeta',
 ]
```

kmdalton:
it's probably okay to merge this if we can't raise @dermen. it don't think it should break any of his workflows.

kmdalton (approved):
this is a sensible extension that enables new machine learning applications.

kmdalton:
@LuisA92 -- I think you should feel free to merge this if you're confident in it. It looks good / safe to me.

PR #337, opened 2026-04-07
Wavelength support for mtz files

kmdalton:
This is a proposal for supporting wavelength in mtz files. It 
 - Adds a wavelength attribute to rs.DataSet and
 - Reads / writes wavelength to mtz files in the io submodule

One downside to this approach is that the mtz file spec actually supports multiple datasets with, potentially, different wavelengths. `rs.from_gemmi` aggregates the info from multiple datasets and stores them in separate columns. From that point, any information about how the columns were structured in the input mtz is lost.

### DIFF
--- a/reciprocalspaceship/io/mtz.py
+++ b/reciprocalspaceship/io/mtz.py
@@ -3,6 +3,7 @@ import gemmi
 from reciprocalspaceship import DataSet
 from reciprocalspaceship.dtypes.base import MTZDtype
 from reciprocalspaceship.utils import in_asu
+import warnings
 
 
 def from_gemmi(gemmi_mtz):
@@ -28,7 +29,21 @@ def from_gemmi(gemmi_mtz):
     -------
     rs.DataSet
     """
-    dataset = DataSet(spacegroup=gemmi_mtz.spacegroup, cell=gemmi_mtz.cell)
+    wavelengths = [d.wavelength for d in gemmi_mtz.datasets]
+    num_uniq = len(set(wavelengths))
+    if num_uniq > 1:
+        message = (
+            "Mtz contains multiple datasets with different wavelengths."
+            f"Choosing first wavelength of available {wavelengths}."
+        )
+        warnings.warn(message)
+    wavelength = wavelengths[0]
+
+    dataset = DataSet(
+        spacegroup=gemmi_mtz.spacegroup, 
+        cell=gemmi_mtz.cell, 
+        wavelength=wavelength,
+    )
 
     # Build up DataSet
     for c in gemmi_mtz.columns:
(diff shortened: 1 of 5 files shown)


reciprocalspaceship/io/mtz.py
```diff
@@ -132,6 +148,13 @@ def to_gemmi(
     mtz.datasets[0].crystal_name = crystal_name
     mtz.datasets[0].dataset_name = dataset_name
 
+    # In gemmi the default wavelength is 0.0 which obviously means no wavelength
```
JBGreisman: Can we remove the "obviously" in the comment?

JBGreisman (changes requested):
I think what's here looks good, but I'd like to see a few tests added and have a question:
1. Tests: Confirm constructor sets the wavelength properly, the property can be accessed, and that a copy() operations preserves the wavelength field.
2. Tests: Round-trip read/write of MTZ sets and preserves wavelength field. I think this should be possible using our synchrotron/monochromatic HEWL dataset that's already a test fixture?
3. Along the lines of writing MTZs, I don't think this is currently plumbed, no? Should we also update the to_gemmi() function to set the relevant field in the gemmi.Mtz object? Is there a particular reason that wasn't implemented?

reciprocalspaceship/io/mtz.py
kmdalton: of course, it's superfluous at best.

kmdalton:
> I think what's here looks good, but I'd like to see a few tests added and have a question:
I can add some tests. Iteration over `DataSet._metadata` in `tests/test_dataset_preserve_attributes.py` catches a lot of potential pathologies, but there is no reason not to add some more explicit tests. 
 
>     1. Tests: Confirm constructor sets the wavelength properly, the property can be accessed, and that a copy() operations preserves the wavelength field.
>     2. Tests: Round-trip read/write of MTZ sets and preserves wavelength field. I think this should be possible using our synchrotron/monochromatic HEWL dataset that's already a test fixture?
I believe these are addressed now. Let me know. 

>     3. Along the lines of writing MTZs, I don't think this is currently plumbed, no? Should we also update the to_gemmi() function to set the relevant field in the gemmi.Mtz object? Is there a particular reason that wasn't implemented?

I vaguely remember discussing multiple crystals per unmerged Mtz a very long time ago. I think we just decided that we didn't want to support that at the time. I don't recall discussing multiple datasets in merged Mtzs, but it would have been a long time ago. I think that these situations are fundamentally different. In the unmerged case, more datasets add rows. In the merged case, columns. Ultimately we should probably revamp the IO to support these configurations. The downside is that we would probably have to stash more Mtz-specific metadata in the DataSet somehow.

JBGreisman (approved):
Thanks for adding the tests. This looks good to me now.

JBGreisman:
I'm mostly remembering this conversation now. In a future (larger) refactor, we can consider how to best account for multi-dataset cases more robustly, which can generalize to `{multi-crystal, multi-wavelength, unmerged data, etc}`.

kmdalton:
sounds good. i'll go ahead and merge this, but we should circle back to these other use cases.

Issue #338, opened 2026-04-08
rs.io.read_dials_stills crashes when there are too many files per job


kmdalton:
The MPI reader for dials stills files crashes if the number of files / amount of data per job is too high. This is because all the files assigned to a rank are aggregated and sent back rank 0. The default COMM has a 2GB limit so it is pretty easy to overload it if you have a lot of files and only a couple ranks. 

`mpi4py` has a workaround for this, which is to use the pickle protocol 5 (https://mpi4py.readthedocs.io/en/stable/mpi4py.util.pkl5.html#examples).

PR #339, opened 2026-04-08
use pkl5 to enable bigger mpi messages

kmdalton:
Addresses #338 to prevent mpi crashes when there are many more files than ranks.


JBGreisman (approved):
looks good to me

Issue #344, opened 2026-06-15
Add support for batch headers


kmdalton:
A user wanted to put `rs` output into Aimless ([see here](https://discourse.rs-station.org/t/output-in-unmerged-batch-mtz/197/3)). This relies on the batch headers being populated which is something we don't do by default. 

More details  in the forum post, but this can be done at the gemmi level with the following pattern

```python
ds = rs.read_mtz(whatever)

 ... #do whatever fancy stuff you want here to sort / modify ds

dataset_id = 0
cell = ds.cell
mtz = ds.to_gemmi()
for b in ds.BATCH.unique():
    batch = gemmi.Mtz.Batch()
    batch.number = b
    batch.cell = cell
    batch.dataset_id = dataset_id
    mtz.batches.append(batch)

mtz_out.write_to_file("combined_sorted_lorentz.mtz")
```

I propose we modify `ds.write_mtz( ... , batch=True)` by adding a `batch` keyword which defaults to `True` where setting `False` retains the old behavior. True will populate the headers using the dataset's BATCH column and cell.

PR #345, opened 2026-06-17
add batching for mtz files

kmdalton:
This attempts to address #344 to enable batching with `dataset.write_mtz`. I could use some eyes on this control flow. 

This changes the signature for `rs.DataSet.write_mtz` adding `batch=True` and `batch_key=None` options. `write_mtz` will generate a naive set of batches in the header if `batch` is truthy. `batch_key` will be inferred and is only explicitly needed of there are multiple columns with BATCH dtype. `batch=False` and `batch_key is not None` will raise a ValueError.

### DIFF
--- a/reciprocalspaceship/dataset.py
+++ b/reciprocalspaceship/dataset.py
@@ -663,6 +663,8 @@ class DataSet(pd.DataFrame):
         project_name="reciprocalspaceship",
         crystal_name="reciprocalspaceship",
         dataset_name="reciprocalspaceship",
+        batch=True,
+        batch_key=None,
     ):
         """
         Write DataSet to MTZ file.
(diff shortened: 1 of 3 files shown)


JBGreisman:
Overall, this looks good to me. I still want to test it out a bit more before fully approving the change. Just wanted to confirm two things first:
1. Are these read properly by aimless?
2. Are these read by phenix in methods such as the table 1 generator, etc that take unmerged data?

kmdalton:
> Overall, this looks good to me. I still want to test it out a bit more before fully approving the change. Just wanted to confirm two things first:
> 
>     1. Are these read properly by aimless?
> 
>     2. Are these read by phenix in methods such as the table 1 generator, etc that take unmerged data?

Valid concerns. I agree we should confirm before merging.

JBGreisman (approved):
looks fine to me -- I think we should just confirm it behaves with phenix / aimless before merging.

kmdalton:
i think this cropped up again over here: https://github.com/rs-station/laue-dials/issues/93

kmdalton:
@JBGreisman -- i know you approved this, but do you think we should go ahead with merging?

